Issue #76, opened 2023-11-19
Handle `.unwrap()`s gracefully


fjarri:
As a temporary measure, there's a number of `unwrap()` throughout the code base. These need to be removed and replaced with error propagation. When it's done, `warn(clippy::unwrap_used)` in `lib.rs` should be enabled (or even changed to `deny`). Also look for `assert!`.

Issue #10, opened 2023-04-12
Can we get rid of `serde(bound)` annotations?


fjarri:
Because of some dark magic in `serde-derive`, we often need to specify explicit bounds when deriving `Serialize`/`Deserialize`. See https://serde.rs/attr-bound.html for the explanation of the problem. Is it possible to get rid of them somehow?

fjarri:
Note: may be possible if https://github.com/ModProg/derive-where/issues/100 is fixed.

fjarri:
Also maybe we can just derive `Serialize` for our empty types and it'll satisfy the bounds?

PR #120, opened 2024-05-22
Update integer crates to pre-release v0.6

### LINKED ISSUE #104: Update to `crypto-bigint=0.6` and `crypto-primes=0.6` when they are released.
As a part of the update:
- we can dissolve the `UintLike`/`UintLikeMod` traits since they're now implemented as `Integer` and `Monty` in `crypto-bigint`.
- we can use `Odd` for the modulus in `PaillierPublicKey` (need to add support for serialization to `Odd` though)
- possibly use `BoxedUint` for integers?

dvdplm:
Fixes #104.

Update `crypto-bigint` to v0.6 and all the fallout of that.

*Note*: should probably not be merged before the final version of `crypto-bigint` and related dependencies are released.

Benchmarks look alright:

```
group                               dp-update-crypto-bigint-104            master-1d210d149dfe
-----                               ---------------------------            -------------------
happy path/KeyGen, 2 parties        1.00    285.4±3.27µs        ? ?/sec    1.05    299.7±2.21µs        ? ?/sec
happy path/KeyRefresh, 2 parties    1.00    155.7±8.83ms        ? ?/sec    1.07   167.3±11.71ms        ? ?/sec
happy path/Presigning, 2 parties    1.00    150.0±2.07ms        ? ?/sec    1.21    181.2±3.13ms        ? ?/sec
happy path/Signing, 2 parties       1.00      2.2±0.05ms        ? ?/sec    1.25      2.8±0.03ms        ? ?/sec
```


synedrion/src/cggmp21/protocols/key_refresh.rs
```diff
@@ -60,10 +60,18 @@ enum KeyRefreshErrorEnum<P: SchemeParams> {
         mu: Randomizer<P::Paillier>,
     },
 }
-
+// TODO: The requirement that this be Serialize/Deserialize is what causes
```
dvdplm, line 62: @fjarri Do you have any advice for me here? [This](https://github.com/entropyxyz/synedrion/pull/120/commits/2af6086b40a6a77927f67cf9a139d5e005406675#diff-94f155822ad48c350d153fcc6840853dc77c75ec9a12a008e2d098913437dbf7R284) along with the requirement of being `Serialize`&`Deserialize` here causes all these bounds to be required and what's worse is they spread out to many places as you can see in this commit. :/

fjarri: I don't think much can be done here, it's caused by weird `serde`'s heuristic for deriving bounds - see #10 and the link to the `serde` issue therein. I find it strange that you have to explicitly specify the bound on `Odd<Uint>` though - is it really required?

dvdplm: Sadly yes, but I think it's simply that `crypto-bigint` is missing the `Serialize`/`Deserialize` impls for `Odd`. It does provide them for `NonZero`, `Wrapping`, `Uint`, `Limb` etc so I don't see why not provide one for `Odd`.

FWIW I ran into similar issues with serde and even tried to dig deeper into the source code. I found it to be a very difficult code base to work with and did not make progress.

Do you think there's any possibility that we can get around having so many types be `Serialize`/`Deserialize` if we provide manual impls for the exact types that we need? In other words: is this a case where we're actually only need to serialize the round messages that are sent on the wire? Or are there so many types of serialization tasks and types that we really need the whole type hierarchy to be serializable?

fjarri: We need to serialize messages and also entities (key shares, aux data and so on), which contain in them all this hierarchy, so I am not sure if writing explicit impls will be any easier. 

Would it help adding the `Serialize` impl for `Odd` in `crypto-bigint`? Seems like an omission there.

dvdplm:
@fjarri Short status update: the PR now compiles and tests pass – _sometimes_. I have two flaky tests (`paillier::encryption::tests::affine_transform` and `paillier::encryption::tests::homomorphic_mul`) still to figure out. And a ton of `TODO`s to address still.

dvdplm:

synedrion/src/sessions/error.rs
```diff
@@ -36,6 +36,7 @@ pub enum Error<Res: ProtocolResult, Verifier> {
 
 /// An error on this party's side.
 /// Can be caused by an incorrect usage, a bug in the implementation, or some environment error.
+
```
fjarri, line 39: A spurious newline

synedrion/src/uint/traits.rs
```diff
@@ -27,289 +14,48 @@ pub(crate) const fn upcast_uint<const N1: usize, const N2: usize>(value: Uint<N1
...
-    fn bits_vartime(&self) -> usize;
-    fn bit(&self, index: usize) -> Choice;
-    fn bit_vartime(&self, index: usize) -> bool;
-    fn neg(&self) -> Self;
-    fn neg_mod(&self, modulus: &Self) -> Self;
-    fn shl_vartime(&self, shift: usize) -> Self;
-    fn shr_vartime(&self, shift: usize) -> Self;
-    fn to_mod(&self, precomputed: &<Self::ModUint as UintModLike>::Precomputed) -> Self::ModUint {
-        Self::ModUint::new(self, precomputed)
+pub trait ToMod: Integer {
```
fjarri, line 17: Do we really need a custom trait here? Can we just use `crypto_bigint::Monty`?

synedrion/src/uint/signed.rs
```diff
@@ -287,42 +364,68 @@ impl<T: UintLike + HasWide> Signed<T> {
...
+            .expect("Just asserted that bound bits is smaller than T's bit precision");
 
         // Sampling in range [0, 2^bound_bits * scale * 2 + 1) and translating to the desired range.
         let positive_bound = scaled_bound
-            .shl_vartime(1)
-            .checked_add(&T::Wide::ONE)
-            .unwrap();
-        let positive_result = T::Wide::random_mod(rng, &NonZero::new(positive_bound).unwrap());
+            .overflowing_shl_vartime(1)
+            .expect("TODO: justify this properly")
```
fjarri, line 384: Probably better leave as `unwrap()` and add a `TODO (#76)` - will be more noticeable when we start fixing that issue.

synedrion/src/uint/signed.rs
```diff
@@ -332,76 +435,447 @@ where
...
+impl<'a, T> core::iter::Sum<&'a Self> for Signed<T>
+where
+    T: Integer + Encoding + crypto_bigint::Bounded,
+{
     fn sum<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
         iter.cloned().sum()
     }
 }
+
+// @reviewers: Implementing `PartialEq`/`Eq` this way disregards the bound when it comes to equality,
```
fjarri, line 568: The reason it had been behaving like that before was that `PartialEq` had been only used in tests, so the comparison of bounds served as an additional assertion. Now that I think about it, perhaps the `PartialEq` impl should be made test-only (regardless of whether we decide to compare bounds or not) to prevent potential misuse.

synedrion/src/uint/signed.rs
```diff
@@ -332,76 +435,447 @@ where
...
+    fn eq(&self, other: &Self) -> bool {
+        self.abs() == other.abs() && self.is_negative().ct_eq(&other.is_negative()).into()
+    }
+}
+impl<T> Eq for Signed<T> where
+    T: crypto_bigint::Bounded + ConditionallySelectable + Encoding + Integer
+{
+}
+
+impl<T> PartialOrd for Signed<T>
```
fjarri, line 585: Do we need `Signed` to be `Ord` anywhere?

synedrion/src/uint.rs
```diff
@@ -4,12 +4,155 @@ mod traits;
...
+
+/// Math functions for generic integers (mostly `Signed`).
+use crypto_bigint::{
+    subtle::{ConditionallySelectable, CtOption},
+    Square,
 };
+use digest::XofReader;
+
+/// Build a `T` integer from an extendable Reader function
+pub(crate) fn uint_from_xof<T>(reader: &mut impl XofReader, modulus: &NonZero<T>) -> T
```
fjarri, line 24: A minor thing, but I would prefer the "node" files to only contain re-exports - makes it easier to navigate. Perhaps we can move the `pow` functions to `uint/pow.rs`, and `uint_from_xof` to `tools/hashing.rs`?

synedrion/src/lib.rs
```diff
@@ -31,7 +31,7 @@ mod uint;
 mod www02;
 
 // Some re-exports to avoid the need for version-matching
-pub use bip32;
+// pub use bip32;
```
fjarri, line 34: Don't forget to remove in the final version

synedrion/src/constructors.rs
```diff
@@ -43,7 +43,6 @@ where
 {
     Session::new::<key_init::Round1<P, Verifier>>(rng, session_id, signer, verifiers, ())
 }
-
```
fjarri, line 46: A spurious deleted newline

fjarri:
Two general considerations:
- I think it is better to have `unwrap()` (and perhaps a `// TODO (#76)`) rather than `expect("TODO: ...")` - will make it easier to fix #76 later. 
- I wonder if it is worth it to retain a custom `Uint` trait to use it an alias for the trait combination we require (`Integer`, `Bounded`, `Serialize` etc) and an umbrella for the `pow_*` functions which are now free-standing.

synedrion/src/uint/signed.rs
dvdplm: We do not, this is indeed not strictly necessary. 

I added it because it made writing tests a bit easier and because it made using `Signed` values more like a "normal" number type.

Do you think I should remove it?

synedrion/src/uint/signed.rs
dvdplm: This is a "note to self" kind of `TODO`, one I intend to fix in this PR before merging; I missed this one before asking for a review, apologies!

I think `expect("why this will never (ever) happen");` is a good pattern generally, because it really forces the author to pause and reflect on why it is ok to unwrap.

synedrion/src/uint/signed.rs
dvdplm: `PartialEq` is definitely used, e.g. [here](https://github.com/dvdplm/synedrion/blob/b02782ad9c987d45ffba20ba839ad25b300085fc/synedrion/src/cggmp21/sigma/fac.rs#L161) so we have to make a call here. The conservative option seems to back out my addition here and revert back to deriving it. Works for you?

dvdplm:
> * I think it is better to have `unwrap()` (and perhaps a `// TODO (#76)`) rather than `expect("TODO: ...")` - will make it easier to fix

I strongly agree with #76. Especially for a library it's a bit rude to panic and take down somebody else's app. That said I also think there are a number of places where unwrapping (or `expect`ing) is warranted, e.g. when we *know* that a `T::Wide` cannot overflow.

How is using `expect` making it harder to work on #76? That issue is going to be a big undertaking regardless I think...

synedrion/src/uint/traits.rs
dvdplm: This is a convenience because type inference is being wonky. Take this example with the trait:

```rust
        let inv_totient = totient
            .as_ref()
            .to_montgomery(public_key.precomputed_modulus())
            .invert()
            .expect("Bla bla");

```

…and without the trait:

```rust
        let inv_totient = <<P as PaillierParams>::UintMod as Monty>::new(
            totient.into_inner(),
            <P::UintMod as Monty>::Params::clone(public_key.precomputed_modulus()),
        )
        .invert()
        .expect("Bla bla");

```

Certainly doable, but readability suffers!

fjarri:
> How is using expect making it harder to work on https://github.com/entropyxyz/synedrion/issues/76? That issue is going to be a big undertaking regardless I think...

If we have an `unwrap()` there, we just enable the `clippy::unwrap_used` lint, and fix the compilation errors. With `expect("TODO: ` we need to specifically look for those entries as well. Also some of those `expect`s may need to be converted into error propagation.

synedrion/src/uint/signed.rs
fjarri: Yes, I think it is better to keep the strict behavior unless we specifically need to relax it.

fjarri:
Sorry, I just realized I contradicted myself in two of the comments regarding the wrapping traits for `Integer`/`Monty`. `ToMod` seems justified to keep, and if the power methods can be put in it, it may simplify all the `pow` usage where we pass the bound explicitly.

dvdplm:
> > How is using expect making it harder to work on #76? That issue is going to be a big undertaking regardless I think...
> 
> If we have an `unwrap()` there, we just enable the `clippy::unwrap_used` lint, and fix the compilation errors. With `expect("TODO: ` we need to specifically look for those entries as well. Also some of those `expect`s may need to be converted into error propagation.

There is also `clippy::expect_used` for that, but I see your point. In this PR, the `unwrap`s that I have turned into `expect`s are now all properly justified and I think they are all legitimate, that is they are of the kind that could/should probably remain `expect`s even after #76.

dvdplm:
> `ToMod` seems justified to keep, and if the power methods can be put in it, it may simplify all the `pow` usage where we pass the bound explicitly.

Not sure I understand what you mean here:

- I named it `ToMod` trait because the code on master calls the method `to_mod`, but I think `to_montgomery` is a more accurate name given it converts a `Uint` to montgomery form, so I renamed the trait `ToMontgomery`.
- I think you're suggesting we can make it into a more general "utility" trait, or "umbrella" that also contains the various flavours of exponentiation code, i.e. the code from the [`uint::pow`](https://github.com/entropyxyz/synedrion/pull/120/files#diff-6ac5f2297e4ac6cf3545c011a496699b8dc9ff17d5f0908d41dc3b3ef6eb24be) module. This works fine, the code can be move over almost verbatim.
- However the name should change now, given it does two things (convert to monty form and also exponentiation). So perhaps two traits is better? Keep `ToMontgomery` and add a `PowUtils` or something similar?
- …but if we add a new trait, what did we gain exactly vs having a `uint::pow`module? Arguably not much.
- Trait methods cannot have visibility qualifiers, so no `pub(crate)` or private methods. Not a big deal, but it's a downside.
- In another comment you mentioned adding an umbrella trait but I didn't quite follow what you had in mind there, do you mind elaborating a bit?

My advice is to try to keep this PR as simple as possible and keep the `pow*` methods in their new `uint::pow` module for the time being. I also like smaller traits, and the `ToMontgomery` trait is small&simple.

synedrion/src/uint/signed.rs
dvdplm: Done!

fjarri:
> …but if we add a new trait, what did we gain exactly vs having a uint::powmodule? 

Basically the end goal for me is to be able to write `x.pow_signed(y)` with `y: Signed` without having to explicitly specify the bound as it's done now (so you have to write `x.pow_signed(y, y.bound())`). I don't mind merging the current PR as is and experiment with it separately. 

For merging, could you rebase is to the current `master` and, if you want a merge commit, clean up/squash the history? Or if you don't want to bother with that, I can just squash merge.

dvdplm:
> For merging, could you rebase is to the current `master` and, if you want a merge commit, clean up/squash the history? Or if you don't want to bother with that, I can just squash merge.

I don't have strong opinions about such things! A squash merge is fine by me. :)

Do note however that we still need a new release of `bip32` and `hashing-serializer` is still a pre-release.

synedrion/Cargo.toml
```diff
@@ -10,23 +10,23 @@ readme = "README.md"
...
 secrecy = { version = "0.9.0-pre.0", default-features = false, features = ["serde"] }
 zeroize = { version = "1.8", default-features = false, features = ["alloc", "zeroize_derive"] }
-bip32 = { version = "0.5.2", default-features = false, features = ["alloc", "secp256k1"] }
+bip32 = { version = "0.5.2", default-features = false, features = ["alloc", "secp256k1", "k256"] }
 
 # Note: `alloc` is needed for `crytpto-bigint`'s dependency `serdect` to be able
 # to serialize Uints in human-readable formats.
-crypto-bigint = { version = "0.5.3", default-features = false, features = ["serde", "alloc"] }
-crypto-primes = { version = "0.5", default-features = false }
+crypto-bigint = { version = "0.6.0-rc.3", features = ["serde", "alloc", "rand_core"] }
```
fjarri, line 28: I think the `rand_core` here may be the thing that's breaking the WASM build in CI

fjarri:
I would rather keep `hashing-serializer` as `.pre` as long as it depends on the `.pre` version of `digest`. As for `bip32`, it is an unfortunate situation where we have a bunch of pre-release version of RustCrypto crates all hinging on `crypto-bigint` being pre-release. I'll probe Tony's intentions on finally making a release, it's been more than a year already.

synedrion/Cargo.toml
```diff
@@ -10,34 +10,38 @@ readme = "README.md"
...
 
 [dependencies]
-signature = { version = "2", default-features = false, features = ["alloc"] }
-k256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "arithmetic"] }
-rand_core = { version = "0.6.4", default-features = false }
-sha2 = { version = "0.10", default-features = false }
-sha3 = { version = "0.10", default-features = false }
-digest = { version = "0.10", default-features = false, features = ["alloc"]}
+signature = { version = "2.3.0-pre.4", default-features = false, features = ["alloc"] }
+k256 = {git = "https://github.com/RustCrypto/elliptic-curves", default-features = false, features = ["ecdsa", "arithmetic"]}
```
fjarri, line 14: I would really like to merge with this pointing to a published version, but Tony seems to be waiting for something. I would rather not keep this PR open any longer, but I think we should at least point to a specific commit here.

synedrion/Cargo.toml
fjarri: Why do we even need this feature outside of `dev-dependencies`? What part of the code uses it?

synedrion/Cargo.toml
fjarri: I'll remind him again.

synedrion/src/paillier/encryption.rs
```diff
@@ -39,11 +41,28 @@ impl<P: PaillierParams> RandomizerMod<P> {
     }
 
     pub fn pow_signed(&self, exponent: &Signed<P::Uint>) -> Self {
-        Self(self.0.pow_signed(exponent))
+        let abs_exponent = exponent.abs();
```
fjarri, line 44: Can we use functions from `uint/pow.rs` here?

synedrion/src/paillier/keys.rs
```diff
@@ -37,48 +48,74 @@ impl<P: PaillierParams> SecretKeyPaillier<P> {
...
         let public_key = public_key.to_precomputed();
 
         let inv_totient = totient
-            .as_ref()
-            .to_mod(public_key.precomputed_modulus())
+            .into_inner()
+            .to_montgomery(public_key.precomputed_modulus())
             .invert()
-            .unwrap();
+            .expect("The modulus is pq. ϕ(pq) = (p-1)(q-1) is invertible mod pq because nor (p-1) or (q-1) share factors with pq.");
```
fjarri, line 79: nitpick: I think it's "neither ... nor"

synedrion/src/paillier/keys.rs
```diff
@@ -37,48 +48,74 @@ impl<P: PaillierParams> SecretKeyPaillier<P> {
...
+        .expect("We assume `P::MODULUS_BITS` is properly configured");
 
-        let inv_p_mod_q = self.p.to_mod(&precomputed_mod_q).invert().unwrap();
-        let inv_q_mod_p = self.q.to_mod(&precomputed_mod_p).invert().unwrap();
+        let inv_p_mod_q = self
+            .p
+            .clone()
+            .to_montgomery(&precomputed_mod_q)
+            .invert()
+            .expect("All non-zero integers have a multiplicative inverse mod a prime");
```
fjarri, line 95: Technically, non-zero mod that prime, but since p != q, it amounts to the same thing.

synedrion/src/uint/signed.rs
fjarri: If it's just for tests, perhaps decorate it with `#[cfg(test)]`

synedrion/Cargo.toml
dvdplm: 🙏

synedrion/src/paillier/encryption.rs
dvdplm: I did some work on more ergonomic exponentiation [here](https://github.com/entropyxyz/synedrion/pull/145) but didn't want to make this PR noisier than it already is so that PR builds upon this one.

synedrion/src/paillier/keys.rs
dvdplm: True! tyvm!

fjarri:
`k256` and `ecdsa` have been bumped, `bip32` not yet. @dvdplm, what's your opinion, should we merge this, and add an issue for bumping the deps to stable versions (which we will only be able to resolve after Rust 1.83 hits, so end of November)? Or wait for two more months?

dvdplm:
I think we’re fine to merge it and deal with the crate versions in subsequent PRs. I don’t love it and I don’t think we should make a habit out of it but I think it’s ok to make an exception here. 

Let’s make sure we have a tag at the commit before merging though so downstream can use something stable.

dvdplm:
@fjarri `bip32` is now updated. Should I merge?

fjarri:
Yep, all looks good to me, finally all the deps are on crates.io. I'm going to squash merge.

Issue #118, opened 2024-05-08
Access to MessageType (and SignedMessage?)


tmpfs:
For usage in real-world scenarios it's necessary to determine whether a message is broadcast or direct to route correctly but `CombinedMessage` does not expose this information as `SignedMessage` is private so I cannot call `SignedMessage::message_type()`.

I can't think of a good reason why it needs to be private, are you ok to expose this?

Will submit a PR.

fjarri:
The current approach is that all messages are considered to be direct ones to simplify the calling code. What platform are you running on that gives you access to broadcasting?

tmpfs:
Hi @fjarri ,

I am using an E2EE relay server (tunnels messages over the noise protocol) but is able to broadcast and send directly. It would really help me if I could call `message_type()` to be able to route messages appropriately.

fjarri:
It is not quite as simple as that. Currently `Session::make_message()` takes a `destination`, so simply exposing the message type will just make the API weird, because the whole structure of the public methods assumes direct messages only. I think a better approach is to have a public `Session` method to tell you if the current round's message is a pure broadcast, and then another method to request it (possibly the same method returning an `Option`). Then you will have to remember not to call `make_message()` in that round to avoid duplicates. I can't immediately think of a way to prevent that on the type level. 

I wonder though, how critical it is to you to use broadcasts instead of direct messages when possible? Have you done any benchmarking? Not to say it's a bad idea, I just have absolutely no intuition on how big the benefits of it are. (In case you're worried that `make_message()` does extra work when there's only a broadcast - it doesn't, the broadcast, if any, is prepared when the round is started and then just copied)

tmpfs:
Oh, I must be thinking about this wrong then. I am upgrading from a GG20 implementation which only yielded a single broadcast message to be dispatched to all other participants. But IIUC `Session::message_destinations` will always yield n-1 (or t-1 for the threshold protocols) messages regardless of whether its a direct, broadcast or combination round, is that correct?

If so, I will just route everything over the direct channels and ignore the broadcast channel on the network transport.

fjarri:
> But IIUC Session::message_destinations will always yield n-1 (or t-1 for the threshold protocols) messages regardless of whether its a direct, broadcast or combination round, is that correct?

That's right (in KeyResharing case, which is a little special, it will be `n` or `n-1` messages, where `n` is the number of new share holders, `-1` if the node is both an old and a new share holder).

[...]

PR #127, opened 2024-06-23
Session/Message API improvements

### LINKED ISSUE #123: It would be nice if `CombinedMessage` exposed the `SessionId` of the message
To do distributed key generation for t of n signing, we are running the key init, reshare and aux gen protocols one after the other using the same channels for parties to send protocol messages to each other.

I have had the issue that sometimes we get a message intended for the next session whilst we are still waiting for a message from different party on the current session. Usually its the first message from the aux gen protocol that comes whilst we are still waiting for the last message of the reshare protocol from another party.

When this happens, the session ends with an error as the session id for the message is not correct. To stop this happening i am sending the session id together with the protocol message in order to be able to able to check if it is for the current session before passing it to the session for processing.

Since `SignedMessage` already contains the `SessionId`, its is a bit inefficient to have to send a second copy of it together with the message - it would be nice if synedrion provided a method to get it.

fjarri:
- Rename `CombinedMessage` to `MessageBundle`
- Expose `MessageBundle::session_id()`, `round()`, `is_echo()`; fixes #123
- `SessionId` is now public

TODO:
- Take `SessionId` instead of `shared_randomness` in `Session::new()` 
- Deal with #118 somehow? Add a `broadcast_only() -> Option<SignedMessage>`?


fjarri:
Deleted by mistake.

PR #130, opened 2024-07-09
Add serde support for ThresholdKeyShare.

tmpfs:
Requires additional Ord trait bounds.


tmpfs:
Hi @fjarri , I would like to be able to send `ThresholdKeyShare` across the JS/Webassembly boundary so we can encrypt the key share and store it on disc.

The easiest way to do this is if `ThresholdKeyShare` (like `KeyShare`) implements the serde traits.

Hope this is ok 🙏

fjarri:
Thanks, I added serialization to `KeyShare`, but forgot about their threshold counterparts.

fjarri:
Codecov uploader acting out, otherwise the tests pass.

PR #131, opened 2024-07-11
Add child key derivation to `ThresholdKeyShare`

fjarri:
At the moment it's just a simple hash-based derivation, not BIP32, since the `bip32` package's API is somewhat limited (a private key cannot be created from an existing secret share). Still, it works as intended:
- the full verifying key can be obtained from any parent key share, given the seed
- the full verifying key of child shares does not change after resharing of the parent shares

This means that BIP32, which works on the same principle, will work as well.


fjarri:
Merging this variant now while waiting on `bip32` to add the required API.

PR #139, opened 2024-07-26
Add `entropyxyz` organization username to allowlist.

vitropy:
It is unlikely that users belonging to the `entropyxyz` GitHub Organization are required to sign the CLA. To automatically allow any Organization member to bypass the requirements to sign the CLA, merge this PR.

PR #140, opened 2024-07-31
Bump `bincode` and `secrecy` for `no_std` builds

HCastano:
I've been strugging to use Synedrion as a dependency in a Substrate pallet due to one of
its dependencies pulling in `std` transitively, as mentioned in https://github.com/entropyxyz/entropy-core/issues/961.

Through a bit of digging I found that the `bincode` and `secrecy` dependencies were
pulling in `serde` without disable default features, causing `std` to be pulled in.

I've updated them to the latest releases which use better feature management to allow
`serde` to not be pulled in with default features. Note that the latest releases are
technically pre-releases.

### DIFF
--- a/synedrion/src/curve/arithmetic.rs
+++ b/synedrion/src/curve/arithmetic.rs
@@ -24,7 +24,7 @@ use k256::{
     Secp256k1,
 };
 use rand_core::CryptoRngCore;
-use secrecy::{CloneableSecret, SecretBox, SerializableSecret};
+use secrecy::{CloneableSecret, SerializableSecret};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use zeroize::DefaultIsZeroes;
 
@@ -163,8 +163,6 @@ impl<'de> Deserialize<'de> for Scalar {
 }
 
 impl DefaultIsZeroes for Scalar {}
-
-// impl DebugSecret for Scalar {}
 
 impl CloneableSecret for Scalar {}
 
--- a/synedrion/src/curve/arithmetic.rs
+++ b/synedrion/src/curve/arithmetic.rs
@@ -24,7 +24,7 @@ use k256::{
     Secp256k1,
 };
 use rand_core::CryptoRngCore;
-use secrecy::{CloneableSecret, DebugSecret, SerializableSecret};
+use secrecy::{CloneableSecret, SecretBox, SerializableSecret};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use zeroize::DefaultIsZeroes;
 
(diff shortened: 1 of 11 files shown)


synedrion/src/sessions/type_erased.rs
```diff
@@ -18,15 +18,16 @@ use crate::rounds::{
 };
 
 pub(crate) fn serialize_message(message: &impl Serialize) -> Result<Box<[u8]>, LocalError> {
-    bincode::serialize(message)
+    bincode::serde::encode_to_vec(message, bincode::config::legacy())
```
HCastano, line 21: So I wasn't sure if we actually needed to care about compatibility with `bincode@1.x` here, so I went with the legacy encoding. If we don't care we can use `bincode::config::standard()` instead

HCastano:
@vitropy looks like I still need to sign the CLA even though I'm part of the Entropy org. I thought this wasn't supposed to be the case after #139 🤔

synedrion/src/sessions/type_erased.rs
fjarri: Not really, it just needs to be consistent with the decoding counterpart.

synedrion/src/paillier/keys.rs
```diff
@@ -12,20 +11,12 @@ use crate::uint::{
     RandomPrimeWithRng, Retrieve, Signed, UintLike, UintModLike,
 };
 
-#[derive(Clone, Serialize, Deserialize, ZeroizeOnDrop)]
+#[derive(Clone, Debug, Serialize, Deserialize, ZeroizeOnDrop)]
```
fjarri, line 14: Does it somehow obscure the secret values in the output, the way the previous implementation did?

HCastano: Yeah, it doesn't log the inner value. See the [implementation on `SecretBox` here](https://docs.rs/secrecy/0.9.0-pre.0/src/secrecy/lib.rs.html#143).

HCastano:

synedrion/src/paillier/keys.rs
fjarri: But `p` and `q` are not `SecretBox`-ed.

HCastano: Oh sorry, you're right. I thought they were `SecretBox`-ed for some reason. I've added a manual `Debug` implementation that matches the `DebugSecret` one.

One other option here would be to `SecretBox` `p` and `q`, but it would add bit of noise of this PR. Wdyt?

synedrion/src/curve/arithmetic.rs
```diff
@@ -164,8 +164,6 @@ impl<'de> Deserialize<'de> for Scalar {
 
 impl DefaultIsZeroes for Scalar {}
 
-impl DebugSecret for Scalar {}
```
HCastano, line 167: I assume we don't need a manual implementation of `DebugSecret` anymore like for `SecretKeyPaillier` since a) this already had a derived `Debug` implementation and b) anybody that's using for secret values these can wrap them in a `SecretBox` and use that `Debug` implementation

PR #141, opened 2024-08-03
Compile some deps with O3 level optimizations

dvdplm:
This PR makes a select number of dependencies compile with the full set of optimizations even in debug mode. Given that dependencies are not often updated, the impact on the dev cycle is pretty small.

On the other hand, the speedup when running the test suite is also pretty small (less than 10% on my M3 CPU).

With O3 opt:

`cargo t  90.81s user 1.04s system 227% cpu 40.331 total`

Without O3 opt:

`cargo t  99.73s user 0.63s system 227% cpu 44.128 total`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,3 +3,16 @@ members = [
     'synedrion',
 ]
 resolver = "2"
+
+[profile.dev.package."crypto-bigint"]
+opt-level = 3
+[profile.dev.package."crypto-primes"]
+opt-level = 3
+[profile.dev.package."digest"]
+opt-level = 3
+[profile.dev.package."k256"]
+opt-level = 3
+[profile.dev.package."sha2"]
+opt-level = 3
+[profile.dev.package."sha3"]
+opt-level = 3


fjarri:
So the goal of this change is the small speed-up in debug mode? Maybe we can just enable -O3 in tests for the whole crate which makes a significantly greater effect on test times?

dvdplm:
We could also do that, and that would speed things up much more of course. The downside is that all tests are slow to build all the time, which might be annoying when working on single test cases and there’s a lot of recompiles. 

When I run the full test suite I usually run with `—release` so to me this isn’t so much about that particular part. 

Only having (select) dependencies build with O3 means that the code we work on ourselves is still quick to build by default, which I think is good. 

Another downside with this PR that I didn’t mention is that O3 strips debugging symbols, so if we ever need to use a stepping debugger, having a mix of crates built with different optimization levels is not going to work great.

I didn’t spend a lot of time on figuring out which dependencies are “heavy” so it’s possible that this initial set could be improved.

No strong feelings about this and if you’re hesitant I’m on closing it! Your call. :)

fjarri:
> Another downside with this PR that I didn’t mention is that O3 strips debugging symbols

Does it? I thought this setting would only override the `opt-level` setting of the `dev` profile without affecting other settings, in particular `strip`.

dvdplm:
> Does it? I thought this setting would only override the `opt-level` setting of the `dev` profile without affecting other settings, in particular `strip`.

Now that I re-read this I am actually not sure. I can easily imagine that stepping through code that is partially O3-optimized and partially not would probably look very odd. Not an important downside I think.

dvdplm:
I ran more benchmarks today and the slow tests are actually just as slow on this PR as they are on master. :/

I even tried setting 

```toml
[profile.dev.package."*"]
opt-level = 3
```

…so that *all* dependencies are built with O3 optimizations. The speedup is negligible. I'll close this for now.

PR #143, opened 2024-08-20
Don't clone when preparing Echo broadcast messages

dvdplm:
Just a small optimization. Given that bincode does not encode struct names anyway, we can get away with serializing a raw `Box` and do the serialization work on references instead.

### DIFF
--- a/synedrion/src/sessions/echo.rs
+++ b/synedrion/src/sessions/echo.rs
@@ -55,16 +55,14 @@ where
         &self.destinations
     }
 
-    pub fn make_broadcast(&self) -> Box<[u8]> {
-        let message = Message {
-            broadcasts: self
-                .broadcasts
-                .clone()
-                .into_iter()
-                .map(|(idx, msg)| (idx, msg.into_unverified()))
-                .collect(),
-        };
-        serialize_message(&message).unwrap()
+    /// Serialize an [`EchoRound`] into a bincode encoded [`Message`].
+    pub fn make_broadcast(&self) -> Result<Box<[u8]>, LocalError> {
+        let message: Box<[(_, _)]> = self
+            .broadcasts
+            .iter()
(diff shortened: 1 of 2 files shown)


synedrion/src/sessions/echo.rs
```diff
@@ -55,16 +55,14 @@ where
...
-                .broadcasts
-                .clone()
-                .into_iter()
-                .map(|(idx, msg)| (idx, msg.into_unverified()))
-                .collect(),
-        };
-        serialize_message(&message).unwrap()
+    /// Serialize an [`EchoRound`] into a bincode encoded [`Message`].
+    pub fn make_broadcast(&self) -> Result<Box<[u8]>, LocalError> {
+        let message: Box<[(_, _)]> = self
```
fjarri, line 60: Serializing a different type from what's later deserialized seems potentially error-prone. Perhaps we can instead have `Message::broadcast` be a vector of references?

dvdplm: With bincode you don't really get type safety anyway and a serialized payload can be deserialized into any type that "fits" the bincode layout. It's really up to us to ensure that we deserialize into a `Message`. The type that was used during serialization is effectively erased during serialization.

Making `Message` be `struct Message { broadcasts: Vec<(&'a I, &'a SignedMessage)> }` sounds tricky: when we deserialize into a `Message` we'd need to keep the bincode buffer that owns the data around for `'a` which is hard. We'd have to have a `MessageOwned` type or some such to make it reasonably ergonomic. Not worth it imo.

fjarri: It's true that it works now, but there may be problems if we want to serialize it in some other format that does contain field names, or perhaps if the structure of the `Message` is changed. Honestly, I'd rather keep the `clone()` and not introduce implicit dependencies in the code.

Issue #60, opened 2023-11-18
Group similar methods using traits


fjarri:
Consider `RPParamsMod` that has three methods `commit()`, `commit_wide()` and `commit_xwide()` which do the same thing, but with different types of integers as arguments. Instead, we can have a `Commit` trait with different associated types. The same applies e.g. for `Ciphertext` constructors.

fjarri:
Doesn't seem to be possible within the Rust type system since e.g. `PaillierParams::Uint` and `PaillierParams::WideUint` may be the same type.

PR #144, opened 2024-08-23
Use `SecretBox` for all secret values

### LINKED ISSUE #77: Add zeroization to secret values
Need to go through structs and wrap secret fields in a `Secret`-like wrapper (see e.g. https://crates.io/crates/secrecy), or add `ZeroizeOnDrop` derivation if all the fields are secret. Check, specifically:
- structs in `protocols/common.rs`
- structs in specific protocols in `protocols`
- structs in the `paillier` module

dvdplm:
Wrap more secrets in `SecretBox`.

Keeping this PR as a draft, given it depends on a pre-release candidate for `secrecy` (see [this issue](https://github.com/iqlusioninc/crates/issues/1208)).

 When `crypto-bigint` v0.6 is released we can address one more `TODO` ([this one](https://github.com/dvdplm/synedrion/blob/a0e96237dbe75a0fa28dc1c28fefface0eb5b143/synedrion/src/paillier/keys.rs#L176)). 

Fixes #77


synedrion/src/cggmp21/protocols/presigning.rs
```diff
@@ -295,12 +301,12 @@ pub struct Round2Message<P: SchemeParams> {
...
-    r: Randomizer<P::Paillier>,                          // TODO (#77): secret
-    s: Randomizer<P::Paillier>,                          // TODO (#77): secret
-    hat_r: Randomizer<P::Paillier>,                      // TODO (#77): secret
-    hat_s: Randomizer<P::Paillier>,                      // TODO (#77): secret
+    beta: SecretBox<Signed<<P::Paillier as PaillierParams>::Uint>>,
+    hat_beta: SecretBox<Signed<<P::Paillier as PaillierParams>::Uint>>,
+    r: SecretBox<Randomizer<P::Paillier>>,
+    s: SecretBox<Randomizer<P::Paillier>>,
+    hat_r: SecretBox<Randomizer<P::Paillier>>,
+    hat_s: SecretBox<Randomizer<P::Paillier>>,
```
dvdplm, line 309: This change is what causes most of the noise in the PR.

synedrion/src/cggmp21/protocols/presigning.rs
```diff
@@ -344,25 +350,41 @@ impl<P: SchemeParams, I: Debug + Clone + Ord + Serialize> Round<I> for Round2<P,
...
+            pk,
+            beta.expose_secret(),
+            r.expose_secret().retrieve().secret_box(),
+        );
         let cap_d = &self.all_cap_k[destination] * P::signed_from_scalar(&self.context.gamma)
-            + CiphertextMod::new_with_randomizer_signed(target_pk, &-beta, &s.retrieve());
+            + CiphertextMod::new_with_randomizer_signed(
+                target_pk,
+                &-beta.expose_secret(),
+                s.expose_secret().retrieve().secret_box(),
```
dvdplm, line 373: There are a few of these "expose-modify-wrap-again" that made me wish there was a more powerful API offered by `secrecy` to let me act on the wrapped secret and return the result as a new `SecretBox`, something like `s.transform_inner(|secret| { secret.retrieve() }`. But perhaps the awkwardness is a feature here?

synedrion/src/paillier/encryption.rs
```diff
@@ -113,7 +126,7 @@ impl<P: PaillierParams> CiphertextMod<P> {
     fn new_with_randomizer_inner(
         pk: &PublicKeyPaillierPrecomputed<P>,
         abs_plaintext: &P::Uint,
-        randomizer: &Randomizer<P>,
+        randomizer: SecretBox<Randomizer<P>>,
```
dvdplm, line 129: It felt a bit silly to take a reference to a `Box`, but perhaps it was fine as it was before (the compiler will likely optimize away the double pointer?). Thought?

fjarri: Wouldn't it be better to have `Randomizer` impl `ZeroizeOnDrop`? I think it is always secret.

fjarri: This way you woudn't have to clone everywhere.

dvdplm: It does! See [here](https://github.com/dvdplm/synedrion/blob/28b36e926b5998a20d36fa5f8f1fd937cfa8dab1/synedrion/src/paillier/encryption.rs#L18). And [`RandomizerMod`](https://github.com/dvdplm/synedrion/blob/28b36e926b5998a20d36fa5f8f1fd937cfa8dab1/synedrion/src/paillier/encryption.rs#L38) does as well.

(This was the case before this PR)

In my head I thought `SecretBox`ing values was orthogonal to zeroizing them; the former to try to ensure nothing is left on the stack and the latter to ensure data is zeroed out after use (regardless of stack/heap). But I guess I was wrong about that then. Maybe `ZeroizeOnDrop` is all we need, and then add `Debug` impls to redact the secrets?

fjarri: The way I see it, `SecretBox` is a wrapper for some types that may or may not be secret (e.g. a `Scalar`, or some byte array), so you add the zeroization to them externally if it's necessary. But if the type is already zeroized at all times, no `SecretBox` is needed.

dvdplm: Ok, I reverted the changes that wrapped `Randomizer` and `RandomizerMod` in a `SecretBox` (which coincidentally made the `clone`s go away too).

synedrion/src/cggmp21/params.rs
```diff
@@ -9,8 +9,9 @@ use crate::uint::{
 };
 
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Zeroize)]
```
fjarri, line 14: Strange that it has to be `Zeroize`, I guess the requirement of an overly needy heuristic in a derivation macro?

synedrion/src/uint/signed.rs
```diff
@@ -202,6 +202,10 @@ impl<T: UintLike> Signed<T> {
         T::ModUint::conditional_select(&abs_mod, &-abs_mod, self.is_negative())
     }
 
+    pub fn secret_box(self) -> SecretBox<Signed<T>> {
```
fjarri, line 205: I am not certain about this method, I think its existence sort of provokes unsafe secret handling. Every time it is used we have a secret value being left on the stack.

synedrion/src/cggmp21/protocols/presigning.rs
```diff
@@ -542,14 +555,17 @@ impl<P: SchemeParams, I: Debug + Clone + Ord + Serialize> FinalizableToNextRound
         let cap_delta = cap_gamma * self.context.k;
 
         let alpha_sum: Signed<_> = payloads.values().map(|p| p.alpha).sum();
-        let beta_sum: Signed<_> = artifacts.values().map(|p| p.beta).sum();
+        let beta_sum: Signed<_> = artifacts.values().map(|p| p.beta.expose_secret()).sum();
```
fjarri, line 558: I wonder how far should we go to protect the secrets? In this sum, for example, the safest way to do it is to actually write out an explicit loop mutating the sum inside a `SecretBox`.

synedrion/src/paillier/encryption.rs
```diff
@@ -26,7 +27,10 @@ impl<P: PaillierParams> Randomizer<P> {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, ZeroizeOnDrop)]
+impl<P: PaillierParams + Zeroize> CloneableSecret for Randomizer<P> {}
```
fjarri, line 30: I don't think we need these?

synedrion/src/paillier/keys.rs
```diff
@@ -1,20 +1,22 @@
...
 
 use super::params::PaillierParams;
 use crate::uint::{
     subtle::{Choice, ConditionallySelectable},
     Bounded, CheckedAdd, CheckedSub, HasWide, Integer, Invert, NonZero, PowBoundedExp, RandomMod,
     RandomPrimeWithRng, Retrieve, Signed, UintLike, UintModLike,
 };
+use secrecy::{ExposeSecret, SecretBox};
 
-#[derive(Clone, Serialize, Deserialize, ZeroizeOnDrop)]
```
fjarri, line 14: I thought if we bound `HalfUint: CloneableSecret + SerializableSecret`, we can derive `Clone` and `Seialize`?

synedrion/src/paillier/keys.rs
```diff
@@ -116,31 +151,37 @@ impl<P: PaillierParams> SecretKeyPaillierPrecomputed<P> {
...
+            Signed::new_positive(self.sk.q.expose_secret().into_wide(), P::PRIME_BITS as u32)
+                .expect("The primes in the `SecretKeyPaillier` are 'safe primes' and positive by construction; the bound is assumed to be configured correctly by the user.")
+                .secret_box(),
         )
     }
 
-    pub fn totient(&self) -> &Bounded<P::Uint> {
-        // TODO (#77): must be wrapped in a Secret
-        &self.totient
+    /// Returns Euler's totient function (`φ(n)`) of the modulus, wrapped in a [`SecretBox`].
```
fjarri, line 168: Random thought: #148

synedrion/src/paillier/keys.rs
```diff
@@ -116,31 +151,37 @@ impl<P: PaillierParams> SecretKeyPaillierPrecomputed<P> {
...
+    pub fn totient(&self) -> SecretBox<Bounded<P::Uint>> {
+        self.totient.into()
     }
 
-    /// Returns Euler's totient function of the modulus.
+    /// Returns Euler's totient function (`φ(n)`) of the modulus as a [`NonZero`]
     pub fn totient_nonzero(&self) -> NonZero<P::Uint> {
-        // TODO (#77): must be wrapped in a Secret
-        NonZero::new(*self.totient.as_ref()).unwrap()
+        // TODO (#77): must be wrapped in a Secret – This must await the release of crypto-bigint
```
fjarri, line 175: Is it still the case in 0.6.0-rc3?

synedrion/src/paillier/keys.rs
```diff
@@ -161,17 +202,25 @@ impl<P: PaillierParams> SecretKeyPaillierPrecomputed<P> {
...
-        let q_rem_half = P::HalfUint::try_from_wide(q_rem).unwrap();
+        let mut p_rem = *elem % NonZero::new(self.sk.p.expose_secret().into_wide()).unwrap();
+        let mut q_rem = *elem % NonZero::new(self.sk.q.expose_secret().into_wide()).unwrap();
+        let mut p_rem_half = P::HalfUint::try_from_wide(p_rem).unwrap();
+        let mut q_rem_half = P::HalfUint::try_from_wide(q_rem).unwrap();
 
         let p_rem_mod = p_rem_half.to_mod(self.precomputed_mod_p());
         let q_rem_mod = q_rem_half.to_mod(self.precomputed_mod_q());
+
+        // TODO (#77): zeroize intermediate values
```
fjarri, line 215: Perhaps they can be wrapped in `SecretBox` so that we can let RAII do its job? Or would that make too much boilerplate?

synedrion/src/paillier/keys.rs
```diff
@@ -356,4 +409,34 @@ mod tests {
...
+        let debug_output = format!("Sikrit {:?}", sk);
+        assert_eq!(debug_output, "Sikrit [REDACTED synedrion::paillier::keys::SecretKeyPaillier<synedrion::paillier::params::PaillierTest>]");
+    }
+
+    #[test]
+    fn serialization_and_clone_works() {
+        let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(123456);
+        let sk = SecretKeyPaillier::<PaillierTest>::random(&mut rng);
+
+        // Can't compare SecretKeyPaillier so we serialize and compare output tokens in this awkward way.
```
fjarri, line 426: Actually, should we just implement `Eq` for it? Should be safe, right?

synedrion/src/cggmp21/params.rs
dvdplm: It doesn't, good catch. I thought `PaillierParams` needed to be `Zeroize` though? Like, if I remove
the requirement
[over here](https://github.com/dvdplm/synedrion/blob/dcb95ce721d95c903c4c459b9a7806ffd67138af/synedrion/src/paillier/params.rs#L11)
then it is no longer needed here of course. On the other hand, this means that we need to be careful
to ensure all components of `PaillierParams` that may contain secrets are zeroized properly. I
believe this is the case at the moment.

synedrion/src/uint/signed.rs
dvdplm: I thought that by passing `self` and moving ownership would remove that risk though. How can I
check? Do you think adding `#[inline(always)]` would help?

synedrion/src/cggmp21/protocols/presigning.rs
dvdplm: Yeah I have the same concern and a poor intuition for what "good enough" means here.

By explicit loop, do you mean something like this?

```rust
       let beta_sum =
            artifacts
                .values()
                .fold(SecretBox::new(Box::new(Signed::default())), |mut sum, p| {
                    *sum.expose_secret_mut() = sum.expose_secret().add(p.beta.expose_secret());
                    sum
                });

```

In this case here I'm not so sure it's worth it. One line below we'd `expose_secret` again to add `beta_sum` to `delta`. But perhaps `delta` is a secret too? 😱

synedrion/src/paillier/encryption.rs
dvdplm: You are correct. Tyvm.

synedrion/src/paillier/keys.rs
dvdplm: It doesn't work (sadly) because the `U512`/`U1024` we use to actually instantiate `HalfUint` do not impl those traits (and both the traits and types are foreign so we can't do it here).

synedrion/src/paillier/keys.rs
dvdplm: In v0.6 it is implemented but this PR is against v0.5 so it'll have to wait. :/

synedrion/src/paillier/keys.rs
dvdplm: I think it would be very noisy and difficult to read. 

I wonder if a macro could help us to be able to say, dunno, `secrets_containing_block! { /* lots of secret calculations here, intermediate values included */ }` and write normal code inside the block and be sure that everything inside will be zeroized and return values will be `SecretBox`ed. 

Could be a rather big undertaking though.

synedrion/src/paillier/keys.rs
dvdplm: Yeah, it's probably fine. It requires `expose_secret()` and I wanted to minimize the number of places we do that, but it's probably ok to have `PartialEq` impl'd.

fjarri:
I think this one should go in next, could you rebase it to `master`?

synedrion/src/paillier/ring_pedersen.rs
```diff
@@ -72,36 +73,44 @@ impl<P: PaillierParams> RPParamsMod<P> {
...
-            pow_signed_wide::<P::Uint>(self.base, randomizer) * pow_signed(self.power, secret),
+            pow_signed_wide::<P::Uint>(self.base, randomizer)
+                * pow_signed(self.power, secret.expose_secret()),
         )
     }
 
     pub fn commit_wide(
         &self,
-        secret: &Signed<P::WideUint>,
+        // TODO(dp): @reviewers Question unrelated to the PR, just something I noticed: Why is the
```
fjarri, line 88: I guess it's just a naming issue. Technically, this method is `commit_wide_secret_wide_randomizer()`, and the next one is `commit_normal_secret_xwide_randomizer()`. These are just overrides for the same operations with different types of arguments. This could be made more straightforward if #60 is fixed.

PR #145, opened 2024-08-29
More ergonomic exponentiation

dvdplm:
As discussed [here](https://github.com/entropyxyz/synedrion/pull/120#issuecomment-2310790463), using a `uint::pow` module is a bit awkward. This PR is an attempt to improve on that by providing an `Exponentiable` trait that implements the various `pow_*` methods with `Signed` exponents.

Note: this PR is based upon #120 and should not be merged before that is merged.

Questions for reviewers:
- What's a better name than `Exponentiable`?
- The trait bounds are kept minimal, and additional bounds added as separate `where` clauses where needed. Is this good or just confusing?


dvdplm:
@fjarri Can we merge this one as well now? :)

PR #150, opened 2024-09-25
Add a test for the unhappy path of signing

dvdplm:
I noticed that the test for the signing protocol does not traverse the unhappy path, where signature verification fails and we generate a correctness proof for the verifying signer. So here's a test for that along some more testing utils.


dvdplm:
@fjarri A review on this would be nice, so we can get rid of CI failures due to missing coverage.

fjarri:
This may have to be modified when #151 lands, but it's a good test for now.

PR #152, opened 2024-10-15
Change totient_nonzero() to return SecretBox

dvdplm:
Resolves the outstanding TODO for #77.

### DIFF
--- a/synedrion/src/paillier/keys.rs
+++ b/synedrion/src/paillier/keys.rs
@@ -156,7 +156,7 @@ impl<P: PaillierParams> SecretKeyPaillier<P> {
 
         SecretKeyPaillierPrecomputed {
             sk: self.clone(),
-            totient,
+            totient: Box::new(totient).into(),
             inv_totient,
             inv_modulus,
             inv_p_mod_q,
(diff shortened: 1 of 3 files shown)


synedrion/src/paillier/keys.rs
```diff
@@ -211,13 +211,10 @@ where
...
-    /// Returns Euler's totient function (`φ(n)`) of the modulus as a [`NonZero`]
-    pub fn totient_nonzero(&self) -> NonZero<P::Uint> {
-        // TODO (#77): must be wrapped in a Secret – This must await the release of crypto-bigint
-        // v0.6, because `NonZero` doesn't implement `Zeroize` in prior versions. Without `Zeroize`
-        // we cannot build a `SecretBox`.
-
-        NonZero::new(*self.totient.as_ref()).expect("φ(n) is a positive, non-zero number")
+    /// Returns Euler's totient function (`φ(n)`) of the modulus as a [`NonZero`], wrapped in a [`SecretBox`].
+    pub fn totient_nonzero(&self) -> SecretBox<NonZero<P::Uint>> {
+        Box::new(NonZero::new(*self.totient.as_ref()).expect("φ(n) is a positive, non-zero number"))
```
fjarri, line 216: We should probably keep it in a `SecretBox` starting right at the creation time, along with other secret fields of `SecretKeyPaillierPrecomputed`.

PR #154, opened 2024-10-17
Return all RustCrypto libs but `crypto-bigint` to stable versions

fjarri:
This will make it possible for `entropy-core` to still use our `master` branch. `crypto-bigint` can stay at the pre-release since it is not directly exposed to the user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `FirstRound::Context` renamed to `Inputs`. ([#102])
 - `Payload` and `Artifact` values are hidden in wrapper types where they were previously exposed. ([#102])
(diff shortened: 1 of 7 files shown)


dvdplm:
Code LGTM.

I wonder though if `core` should depend on `master`. Aren't they better off using a tag or a branch? Right now they are using `0.2.0-beta.0` (which I think correspond to `release/v0.2.0-pre.0` but I'm not actually sure) and that seems like the right thing to do.

I suggest we do this:

0. Make sure we know which tag corresponds to `0.2.0-beta.0`
1. Review&merge the open PRs (or reject&close ofc)
2. Merge this PR
3. Cut a new release `0.2.0-pre.1` with (mostly) stable RC crates

dvdplm (approved):

fjarri:
I guess what I meant was that we won't have to maintain two separate branches and backport changes/fixes, and `entropy-core` will be able to depend on the releases made on `master`.

Issue #36, opened 2023-09-15
Implement Presigning/Signing with `O(N)` identification cost


fjarri:
(Figure and section numbers correspond to the reference paper https://eprint.iacr.org/2021/060)

- [ ] Figure 9: ECDSA Pre-Signing (Rounds 1 to 4) (includes `Π^{enc-elg}`, `Π^{aff-g}`, `Π^{aff-p}`, `Π^{log*}`)
- [ ] Figure 10: ECDSA Pre-Signing (Round 5, 6 & Output) (includes `Π^{log*}`, `Π^{elog}`)
- [ ] Figure 13: ECDSA Signing

Errors:
- [ ] Figure 11: Nonce-Reveal Fail (includes `П^{Nth}`)
- [ ] Figure 12: Pseudo-Key Reveal Fail (includes `П^{Nth}`, `Π^{log}`)

Sigma-protocols:
- [ ] `Π^{Nth}` (Section 2.2.2) - no figure but probably the same as Fig. 23?
- [ ] Figure 23: Dlog Equality – `Π^{log}`
- [ ] Figure 24: Dlog with El-Gamal Commitment – `Π^{elog}`
- [ ] Figure 26: Paillier Affine Operation with Paillier Commitment ZK-Proof – `Π^{aff-p}`
- [ ] Figure 27: Range Proof w/ EL-Gamal Commitment – `Π^{enc-elg}`

fjarri:
The new version of the CGGMP paper removes the 6-round version of the Presigning, since the 3-round one now has O(N) complexity too. This issue is superseded by #157

PR #156, opened 2024-10-26
Switch protocol framework to `manul`

### LINKED ISSUE #151: Manulization
Time to move to https://github.com/entropyxyz/manul, which has reached functional parity with the `rounds`/`session` code here.

fjarri:
Fixes #151 

Stacked on top of https://github.com/entropyxyz/manul/pull/72

As discussed in Slack, it's quite tricky to keep Presigning and Signing separate - if an error happens in Signing it requires a bunch of signed messages from Presigning, and it's very awkward to keep and supply them. So this PR merges Presigning and Signing. 

If we ever need a separate Presigning, the 6-round version from the paper will be suitable, since its Signing counterpart does not require us to transfer any Presigning messages. See #36.

Even though the code is still kind of dirty, this PR is so huge I think it should just stop here. The cleanup will be made in subsequent PRs. We didn't introduce any new unfinished parts, all the unimplemented stuff has already been there before.

*Update:* the new version of CGGMP paper resolves this by moving all the required proofs into Presigning, but I would rather not modify this PR any more, and there's also another one stacked on top of it. The Presigning and Signing will be split again in the course of fixing #157


dvdplm:
Fair enough!

Issue #27, opened 2023-04-13
Generalize the library to use any curve


fjarri:
Right now we have `k256` encoded, but it doesn't have to be. The scheme is applicable to any group with hard logarithm. At the very least we can generalize it to anything implementing the necessary traits from RustCrypto stack.

The main problem would be to make sure, as statically as possible, that the chosen Uint size exceeds the order of the curve scalar.

tarcieri:
@fjarri `elliptic_curve::Curve::Uint` will always exceed the order because it's used to represent the order (as `elliptic_curve::Curve::ORDER`)

fjarri:
The problem is that it doesn't just need to exceed it, but be a certain amount of times larger (x2/4/8). Although it may still be possible to declare statically (or use `BoxedUint`)

tarcieri:
You can use `Concat::Output` to get 2X, and if that type also impl's `Concat`, you can do `<<Self as Concat>::Output> as Concat>::Output` to get 4X.

fjarri:
I made a https://crates.io/crates/tiny-curve crate that we can use for tests, to avoid weird bounds in the parameters and cryptic errors when they are chosen incorrectly. Now we can just use, say, a 32-bit curve and 256-bit primes, making tests significantly faster.

PR #160, opened 2024-11-22
Test FacProof with bad Paillier keys

dvdplm:


synedrion/src/paillier/encryption.rs
```diff
@@ -228,9 +230,10 @@ impl<P: PaillierParams> CiphertextMod<P> {
         // where `m` is the plaintext, `C` is the ciphertext,
         // `N` is the Paillier composite modulus,
         // `phi` is the Euler totient of `N`, and `mu = phi^(-1) mod N`.
+        // Note that ϕ(N) is always invertible, because `gcd(ϕ(N), N) = 1` by construction.
 
         // `C^phi mod N^2` may be 0 if `C == N`, which is very unlikely for large `N`.
-        // Note that `C^phi mod N^2 / N < N`, so we can unwrap when converting to `Uint`
+        // Note that `C^phi mod N^2 / N < N`, so we can unwrap when converting to `Uint` <–– (DP) This assumes that `gcd(C, N) = 1`, but why is this guaranteed?
```
fjarri, line 236: Not sure what you mean here, if `y = x mod N^2` then `y < N^2`, which means `y / N < N`. Hhow does GCD factor in?

fjarri:
I think the reason the test passes is that the proof conditions are violated. You're sampling a small `p` and a regular-sized `q`, which makes the resulting `N_0` to be of size `PRIME_BITS + x`, instead of `2 * PRIME_BITS`. The `П_fac` assumes `N0 ~ 2^(2l + eps) ~ 2^(2 * PRIME_BITS)`. The size of `N_0` would be checked prior to `П_fac` (See Fig. 6, AuxInfo/KeyRefresh, Round 3, Step 1).

It is not trivial to make the proof fail while keeping the `N_0` the proper size, because of the `HalfUint` limits on `p` and `q` (which raises the question of how it should be tested). But if you were to change their types, and have, say `q ~ 2^(PRIME_BITS * 2)`, in the proof (see Fig. 28) you'd get `z2 ~ 2^(4l + 2eps) * q`, which would fail the range check `z2 < 2^(l + eps) * sqrt(N0) ~ 2^(3l + 2eps)`. `l = 256`, so the difference would be quite noticeable.

(the references here are to the '21 version of the paper)

dvdplm:
Maybe all we need to do here is adding a check that `N_0` is indeed `2*PRIME_BITS` bits long?

I struggle to come up with an attack scenario where the a wrong-size `N_0` could slip through the previous rounds, but given that the Paillier secret key is deserializable from bytes (without size checks) it is perhaps wise to adopt a defense in depth approach here.

synedrion/src/paillier/encryption.rs
dvdplm: I honestly no longer remember why I wrote that. Pff. :/

dvdplm:
Maybe worth revisiting this once we close #27 and can test with different param sizes. Closing this for now.

Issue #157, opened 2024-11-14
Implement the updates in the CGGMP paper


fjarri:
The paper we use, https://eprint.iacr.org/2021/060, had some updates in Oct 2024 compared to the 2021 version we used previously. We need to find out what they are and implement them.

PR #161, opened 2024-11-25
Updates for CGGMP'24: Paillier/ring-Pedersen level 

fjarri:
Part of #157

- Split Paillier keys and ring-Pedersen secrets, with the underlying `SecretPrimes`/`PublicModulus`. Note that according to the new version of the paper Paillier only needs `mod 3 = 4` primes, while ring-Pedersen needs safe primes.
- Use `Exponentiable` where approriate
- Normalize naming for "wire" and "expanded" structures in `paillier`. Namely: the wire ones have the "Wire" postfix, the expanded ones have no postfixes. The conversion methods are `to_wire()` and `to_precomputed()`.
- Changed the type of various Uint bit size arguments to `u32` to match `crypto-bigint`

The new `Secret` will be properly used throughout the codebase in a follow-up PR, this one is getting pretty big already.

### DIFF
--- a/synedrion/src/cggmp21/sigma/prm.rs
+++ b/synedrion/src/cggmp21/sigma/prm.rs
@@ -29,7 +29,7 @@ struct PrmSecret<P: SchemeParams>(Vec<Bounded<<P::Paillier as PaillierParams>::U
 impl<P: SchemeParams> PrmSecret<P> {
     fn random(rng: &mut impl CryptoRngCore, secret: &RPSecret<P::Paillier>) -> Self {
         let secret = (0..P::SECURITY_PARAMETER)
-            .map(|_| secret.random_field_elem(rng))
+            .map(|_| secret.random_residue_mod_totient(rng))
             .collect();
         Self(secret)
     }
--- a/synedrion/src/cggmp21/sigma/prm.rs
+++ b/synedrion/src/cggmp21/sigma/prm.rs
@@ -90,7 +90,7 @@ impl<P: SchemeParams> PrmProof<P> {
         setup: &RPParams<P::Paillier>,
         aux: &impl Hashable,
     ) -> Self {
-        // TODO: check that secret.public_modulus == setup.public_modulus?
+        debug_assert!(&secret.modulus() == setup.modulus());
         let proof_secret = PrmSecret::<P>::random(rng, secret);
         let commitment = PrmCommitment::new(&proof_secret, setup.base());
 
--- a/synedrion/src/cggmp21/sigma/prm.rs
+++ b/synedrion/src/cggmp21/sigma/prm.rs
@@ -12,7 +12,7 @@ use serde::{Deserialize, Serialize};
 
 use super::super::SchemeParams;
 use crate::{
-    paillier::{PaillierParams, RPParamsMod, RPSecret},
+    paillier::{PaillierParams, RPParams, RPSecret},
     tools::hashing::{Chain, Hashable, XofHasher},
     uint::{
         subtle::{Choice, ConditionallySelectable},
@@ -49,15 +49,11 @@ impl<P: SchemeParams> PrmCommitment<P> {
 struct PrmChallenge(Vec<bool>);
 
 impl PrmChallenge {
-    fn new<P: SchemeParams>(
-        commitment: &PrmCommitment<P>,
-        setup: &RPParamsMod<P::Paillier>,
-        aux: &impl Hashable,
-    ) -> Self {
+    fn new<P: SchemeParams>(commitment: &PrmCommitment<P>, setup: &RPParams<P::Paillier>, aux: &impl Hashable) -> Self {
         // TODO: use BitVec here?
         let mut reader = XofHasher::new_with_dst(HASH_TAG)
             .chain(commitment)
-            .chain(&setup.retrieve())
+            .chain(&setup.to_wire())
             .chain(aux)
             .finalize_to_reader();
         let mut bytes = vec![0u8; P::SECURITY_PARAMETER];
@@ -91,7 +87,7 @@ impl<P: SchemeParams> PrmProof<P> {
     pub fn new(
         rng: &mut impl CryptoRngCore,
         secret: &RPSecret<P::Paillier>,
-        setup: &RPParamsMod<P::Paillier>,
+        setup: &RPParams<P::Paillier>,
         aux: &impl Hashable,
     ) -> Self {
         // TODO: check that secret.public_modulus == setup.public_modulus?
@@ -118,7 +114,7 @@ impl<P: SchemeParams> PrmProof<P> {
     }
 
     /// Verify that the proof is correct for a secret corresponding to the given RP parameters.
-    pub fn verify(&self, setup: &RPParamsMod<P::Paillier>, aux: &impl Hashable) -> bool {
+    pub fn verify(&self, setup: &RPParams<P::Paillier>, aux: &impl Hashable) -> bool {
         let monty_params = setup.monty_params_mod_n();
 
         let challenge = PrmChallenge::new(&self.commitment, setup, aux);
@@ -147,7 +143,7 @@ mod tests {
     use super::PrmProof;
     use crate::{
         cggmp21::TestParams,
-        paillier::{RPParamsMod, RPSecret},
+        paillier::{RPParams, RPSecret},
     };
 
     #[test]
@@ -155,7 +151,7 @@ mod tests {
         type Params = TestParams;
 
         let secret = RPSecret::random(&mut OsRng);
-        let setup = RPParamsMod::random_with_secret(&mut OsRng, &secret);
+        let setup = RPParams::random_with_secret(&mut OsRng, &secret);
 
         let aux: &[u8] = b"abcde";
 
--- a/synedrion/src/cggmp21/sigma/prm.rs
+++ b/synedrion/src/cggmp21/sigma/prm.rs
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 
 use super::super::SchemeParams;
 use crate::{
-    paillier::{PaillierParams, RPParamsMod, RPSecret, SecretKeyPaillierPrecomputed},
+    paillier::{PaillierParams, RPParamsMod, RPSecret},
     tools::hashing::{Chain, Hashable, XofHasher},
     uint::{
         subtle::{Choice, ConditionallySelectable},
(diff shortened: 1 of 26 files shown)


synedrion/src/tools/hide_debug.rs
```diff
@@ -33,3 +35,49 @@ impl<T> DerefMut for HideDebug<T> {
         &mut self.0
     }
 }
+
+pub(crate) struct Secret<T: Zeroize>(SecretBox<T>);
```
dvdplm, line 39:
```suggestion
/// Helper wrapper for managing secret values, ensuring that secrets do not end up in logs, are zeroized on drop and are serialized/deserialized properly.
pub(crate) struct Secret<T: Zeroize>(SecretBox<T>);
```

synedrion/src/tools.rs
```diff
@@ -5,7 +5,7 @@ pub(crate) mod hashing;
 mod hide_debug;
 pub(crate) mod sss;
 
-pub(crate) use hide_debug::HideDebug;
+pub(crate) use hide_debug::{HideDebug, Secret};
```
dvdplm, line 8: The module name should probably change now that `Secret` is included.

synedrion/src/paillier/rsa.rs
```diff
@@ -0,0 +1,282 @@
...
+            let r = P::Uint::random_mod(rng, &modulus);
+            let r_m = P::UintMod::new(r, self.monty_params_mod_n.clone());
+            if r_m.invert().is_some().into() {
+                return r_m;
+            }
+        }
+    }
+
+    /// Returns a uniformly chosen quadratic residue modulo $N$, in Montgomery form.
+    pub fn random_square_group_elem(&self, rng: &mut impl CryptoRngCore) -> P::UintMod {
```
dvdplm, line 279:
```suggestion
    pub fn random_quadratic_residue(&self, rng: &mut impl CryptoRngCore) -> P::UintMod {
```
Maybe a better name?

synedrion/src/paillier/rsa.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    /// Returns a uniformly chosen number in range $[0, N)$ such that it is invertible modulo $N$, in Montgomery form.
+    pub fn random_invertible_group_elem(&self, rng: &mut impl CryptoRngCore) -> P::UintMod {
+        let modulus = self.modulus_nonzero();
+        loop {
+            let r = P::Uint::random_mod(rng, &modulus);
+            let r_m = P::UintMod::new(r, self.monty_params_mod_n.clone());
+            if r_m.invert().is_some().into() {
+                return r_m;
+            }
```
dvdplm, line 274: Converting to montgomery form on each iteration is pretty expensive; maybe checking `gcd(r, modulus) == 1` and then convert to montgomery outside the loop is faster?

(OTOH almost all `r` will be invertible so perhaps the loops almost always traversed just once anyway...)

synedrion/src/paillier/rsa.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn new(modulus: PublicModulusWire<P>) -> Self {
+        let odd_modulus = Odd::new(modulus.0).expect("the RSA modulus is odd");
+        let monty_params_mod_n = P::UintMod::new_params_vartime(odd_modulus);
+        Self {
+            modulus,
+            monty_params_mod_n,
+        }
+    }
+
+    pub fn to_wire(&self) -> PublicModulusWire<P> {
```
dvdplm, line 246: Not sure I understand the naming here. Is this "wire" as in "the wire format", i.e. the form of this data that ends up in messages? Or?

synedrion/src/paillier/rsa.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn eq(&self, other: &PublicModulus<P>) -> bool {
+        // Only compare the moduli themselves, not the precomputed constants.
+        self.modulus.eq(&other.modulus)
+    }
+}
+
+impl<P: PaillierParams> Eq for PublicModulus<P> {}
+
+impl<P: PaillierParams> PublicModulus<P> {
+    pub fn new(modulus: PublicModulusWire<P>) -> Self {
```
dvdplm, line 237: Should we adhere to the `crypto-bigint` naming convention and call this `new_vartime`? Or at least document that it is not CT?

synedrion/src/paillier/rsa.rs
```diff
@@ -0,0 +1,282 @@
...
+pub(crate) struct PublicModulusWire<P: PaillierParams>(P::Uint);
+
+impl<P: PaillierParams> PublicModulusWire<P> {
+    fn new(primes: &SecretPrimesWire<P>) -> Self {
+        Self(primes.p.expose_secret().mul_wide(primes.q.expose_secret()))
+    }
+
+    pub fn into_precomputed(self) -> PublicModulus<P> {
+        PublicModulus::new(self)
+    }
```
dvdplm, line 216: For consistency this should probably be named `into_public_modulus`, or `PublicModulus` should be renamed `PrecomputedModulus`. I think I prefer the latter, because the name `PublicModulus` implies that there is something like a "private modulus", which isn't really the case.

synedrion/src/paillier/ring_pedersen.rs
```diff
@@ -1,134 +1,166 @@
...
+            Bounded::new(P::Uint::random_mod(rng, bound.expose_secret()), P::MODULUS_BITS - 2)
+                .expect("totient < N < 2^MODULUS_BITS, so totient / 4 < 2^(MODULUS_BITS - 2)")
+        })
+        .into();
+
+        Self { primes, lambda }
+    }
+
+    pub fn lambda(&self) -> &SecretBox<Bounded<P::Uint>> {
+        &self.lambda
```
dvdplm, line 43: So this surprised me at first, but then I realized there's an `AsRef` impl. I honestly think it'd be more readable if you extracted the `SecretBox` from the `Secret` explicitly.

synedrion/src/paillier/ring_pedersen.rs
```diff
@@ -1,134 +1,166 @@
+/// Implements the Definition 3.3 from the CGGMP'21 paper and related operations.
```
dvdplm, line 1: We should be careful about references now that there are two versions of the paper. How about we use `CGGMP21` for the original version and `CGGMP21/24` for the new version?

Which one is this?

dvdplm (approved):
LGTM.

Left a few comments and nitpicks. At the end of the day I think using "wire" is probably fine, but I don't love it and think that the previous names were mostly ok and quite readable.

synedrion/src/paillier/rsa.rs
fjarri: We will need the result in Montgomery form anyway, so it doesn't go entirely to waste. But I agree, since GCD is available now, it may be faster to use that.

synedrion/src/paillier/rsa.rs
fjarri: Yes, exactly that.

synedrion/src/paillier/ring_pedersen.rs
fjarri: I am careful, this particular definition is 3.3 in both :) But there will be updates once I start with ZK proofs and protocols.

synedrion/src/paillier/ring_pedersen.rs
fjarri: That's being done in #163

synedrion/src/tools.rs
fjarri: It'll be renamed in #163. I moved as much of the `Secret` stuff there, but some still had to be kept because it's just too much work to extract it from the PR

fjarri:
> At the end of the day I think using "wire" is probably fine, but I don't love it and think that the previous names were mostly ok and quite readable.

The problem with the previous names was that the most used ones were also very long and had suffixes that had to be repeated everywhere. The goal was to make it so that one could use `PublicKeyPaillier`/`SecretKeyPaillier` in most places, and where they have to be serialized, they're converted to a "minimized" representation which has a longer name. Also sometimes it was `Precomputed`, and sometimes `Mod`, and the conversion method names were different while they were actually doing the same thing. Whether it's `Wire` or something else is secondary and can be changed.

synedrion/src/cggmp21/sigma/prm.rs
```diff
@@ -52,15 +49,11 @@ impl<P: SchemeParams> PrmCommitment<P> {
...
-        commitment: &PrmCommitment<P>,
-        setup: &RPParamsMod<P::Paillier>,
-        aux: &impl Hashable,
-    ) -> Self {
+    fn new<P: SchemeParams>(commitment: &PrmCommitment<P>, setup: &RPParams<P::Paillier>, aux: &impl Hashable) -> Self {
         // TODO: use BitVec here?
         let mut reader = XofHasher::new_with_dst(HASH_TAG)
             .chain(commitment)
-            .chain(&setup.retrieve())
+            .chain(&setup.to_wire())
```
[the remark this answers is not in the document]
fjarri: That terminology is used specifically for Montgomery representation, while this method converts the setup parameters to a minimized serializable form. Which may or may not involve Montgomery conversion, and I didn't want to imply a connection there.

synedrion/src/tools.rs
fjarri: Moved to `tools::secret`

synedrion/src/paillier/rsa.rs
fjarri: Since it's explicitly public info, I don't think we need to make this distinction

synedrion/src/paillier/rsa.rs
fjarri: Changed to use GCD

synedrion/src/cggmp21/sigma/prm.rs
```diff
@@ -93,22 +86,22 @@ impl<P: SchemeParams> PrmProof<P> {
...
-        sk: &SecretKeyPaillierPrecomputed<P::Paillier>,
-        lambda: &RPSecret<P::Paillier>,
-        setup: &RPParamsMod<P::Paillier>,
+        secret: &RPSecret<P::Paillier>,
+        setup: &RPParams<P::Paillier>,
         aux: &impl Hashable,
     ) -> Self {
-        let proof_secret = PrmSecret::<P>::random(rng, sk);
-        let commitment = PrmCommitment::new(&proof_secret, &setup.base);
+        // TODO: check that secret.public_modulus == setup.public_modulus?
```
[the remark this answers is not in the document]
fjarri: Added a debug assertion

synedrion/src/paillier/rsa.rs
```diff
@@ -0,0 +1,282 @@
...
+pub(crate) struct SecretPrimesWire<P: PaillierParams> {
+    p: Secret<P::HalfUint>,
+    q: Secret<P::HalfUint>,
+}
+
+impl<P: PaillierParams> SecretPrimesWire<P> {
+    /// A single constructor to check the invariants
+    fn new(p: Secret<P::HalfUint>, q: Secret<P::HalfUint>) -> Self {
+        debug_assert!(p.expose_secret().as_ref()[0].0 & 3 == 3);
+        debug_assert!(q.expose_secret().as_ref()[0].0 & 3 == 3);
```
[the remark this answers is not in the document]
fjarri: It's more about consistency, these checks are basically documentation, not really something that we need to check every time. If there's `assert!` in the code, it immediately raises a question about why doesn't this return an error instead.

fjarri:
Merging this for now to free the way for #163, we can work on naming in a separate PR.

PR #163, opened 2024-11-28
Secrets rework

fjarri:
This PR is supposed to be the best effort of hiding secret data. Unfortunately we cannot guarantee that no copies are left on the stack in the dependencies.

The general philosophy of the PR is that `.expose_secret()` should appear as little as possible, and whatever is exposed should be either made public with arithmetic operations or immediately re-wrapped in `Secret`.

- Add `Secret`, a convenience wrapper around `SecretBox` supporting serialization/safe debug/cloning/arithmetic operations
- Wrap all the secret data that was remaining in the open (randomizers, secret shares, secret share changes, randoms in the proofs etc)
- Remove `RandomizerWire` since the Paillier randomizers are not supposed to be serialized or sent. What we send out with proofs are masked randomizers (`x * r^y`), so now they have their `MaskedRandomizer` type.

I am not entirely happy with the proliferation of methods for different widths of integers, but a trait won't work because they technically be the same type.


synedrion/src/tools/secret.rs
```diff
@@ -39,22 +73,349 @@ impl<T: Zeroize> Debug for Secret<T> {
...
+            let mut value = self.expose_secret().clone();
+            let result = value.to_wide();
+            value.zeroize();
+            result
+        })
+    }
+
+    pub fn mul_wide(&self, rhs: &T) -> Secret<T::Wide> {
+        Secret::init_with(|| {
+            let mut value = self.expose_secret().clone();
```
dvdplm, line 92: So this is `mut` so that you can `zeroize` it yeah? I wonder if you can get rid of the `clone` and instead take ownership of the secret by `mem::swap`-ing in a `T::ZERO`?
The method takes `&self` which implies that the `Secret` can still be used afterwards, but I wonder how often we actually do that. Like, would a `mul_wide(self)` actually be a problem for us?

synedrion/src/tools/secret.rs
```diff
@@ -39,22 +73,349 @@ impl<T: Zeroize> Debug for Secret<T> {
...
+    T: Zeroize + Clone + Encoding + Integer + HasWide + ConditionallySelectable + crypto_bigint::Bounded,
+    T::Wide: ConditionallySelectable + crypto_bigint::Bounded + Zeroize,
+{
+    pub fn to_wide(&self) -> Secret<Signed<<T as HasWide>::Wide>> {
+        Secret::init_with(|| {
+            let mut value = *self.expose_secret();
+            let result = value.to_wide();
+            value.zeroize();
+            result
+        })
```
dvdplm, line 111: For my education, can you explain why doing what you're doing here is better than:

```rust
Secret::init_with(||
	self.expose_secret().to_wide()
)
```

synedrion/src/tools/secret.rs
```diff
@@ -39,22 +73,349 @@ impl<T: Zeroize> Debug for Secret<T> {
...
+        result %= other;
+        result
+    }
+}
+
+// Summation
+
+impl<T: Zeroize + Clone + for<'a> AddAssign<&'a T> + Default> core::iter::Sum for Secret<T> {
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.reduce(Add::add).unwrap_or(Secret::init_with(|| T::default()))
```
dvdplm, line 334: Maybe we can impl `Default` for `Secret<T: Default>`?

synedrion/src/tools/secret.rs
```diff
@@ -39,22 +73,349 @@ impl<T: Zeroize> Debug for Secret<T> {
...
+    }
+}
+
+impl<T: Zeroize> Secret<T> {
+    pub fn pow_bounded<V>(&self, exponent: &Bounded<V>) -> Self
+    where
+        T: Exponentiable<V>,
+        V: Integer + crypto_bigint::Bounded + Encoding + ConditionallySelectable,
+    {
+        // TODO: do we need to implement our own windowed exponentiation to hide the secret?
```
dvdplm, line 388: What is it about `pow_[bounded, signed]` that worries you? Isn't the same concern valid for `mul`, `invert` and many other methods?

synedrion/src/uint/signed.rs
```diff
@@ -492,17 +462,6 @@ where
     }
 }
 
-impl<T> Add<&Signed<T>> for Signed<T>
```
dvdplm, line 495: Was this no longer needed?

synedrion/src/uint/signed.rs
```diff
@@ -532,82 +491,37 @@ where
...
-impl<'a, T> core::iter::Sum<&'a Self> for Signed<T>
-where
-    T: Integer + Encoding + crypto_bigint::Bounded,
-{
-    fn sum<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
-        iter.cloned().sum()
-    }
-}
-
-impl<T> PartialOrd for Signed<T>
```
dvdplm, line 587: Not needed or…?

synedrion/src/uint/traits.rs
```diff
@@ -196,11 +215,11 @@ impl HasWide for U2048 {
     fn square_wide(&self) -> Self::Wide {
         self.square_wide().into()
     }
-    fn into_wide(self) -> Self::Wide {
-        (self, Self::ZERO).into()
+    fn to_wide(&self) -> Self::Wide {
+        Uint::concat_mixed(self, &Self::ZERO)
```
dvdplm, line 219: Why do you prefer these lower level calls to the `From` impls? Just do be more explicit?

synedrion/src/paillier/encryption.rs
```diff
@@ -160,40 +171,75 @@ impl<P: PaillierParams> Ciphertext<P> {
...
-    pub fn new_with_randomizer(pk: &PublicKeyPaillier<P>, plaintext: &P::Uint, randomizer: &RandomizerWire<P>) -> Self {
+    pub fn new_with_randomizer(
+        pk: &PublicKeyPaillier<P>,
+        plaintext: &Secret<P::Uint>,
+        randomizer: &Randomizer<P>,
+    ) -> Self {
         Self::new_with_randomizer_inner(pk, plaintext, randomizer, Choice::from(0))
     }
 
+    pub fn new_with_randomizer_bounded(
```
dvdplm, line 182: Maybe `new_with_bounded_randomizer`? Reads a bit better. Same for the signed version.

synedrion/src/paillier/encryption.rs
```diff
@@ -207,33 +253,42 @@ impl<P: PaillierParams> Ciphertext<P> {
...
 
-        let mut result = Signed::new_from_unsigned(
-            P::Uint::conditional_select(&positive_result, &negative_result, is_negative),
-            P::MODULUS_BITS - 1,
-        )
-        .expect("the value is within `[-2^(MODULUS_BITS-1), 2^(MODULUS_BITS-1)]` by construction");
+        // Note that this is in range `[0, N)`
+        let positive_result = self.decrypt(sk);
+
+        // Can't define a `Sub<Secret>` for `Uint`, so have to re-wrap manually.
```
dvdplm, line 281: What's the blocker for that?

synedrion/src/uint/signed.rs
fjarri: Since the code compiles, and `Signed` is `pub (crate)`, seems that yes, it's not needed.

synedrion/src/uint/traits.rs
fjarri: If you're talking about the body of the function, it's because the `From` takes the arguments by value, while `concat_mixed()` takes a reference. So there's less chance of leaving secret stuff on the stack.

synedrion/src/paillier/encryption.rs
fjarri: `new_with_bounded_randomizer` implies that the randomized is `Bounded`, but it's not the case - the distinguishing feature is that the secret is `Bounded`.

synedrion/src/paillier/encryption.rs
fjarri: I would be implementing a foreign trait (`Sub`) for a foreign type (or for a possibly foreign type, something like `impl<T: Zeroize + for<'a> SubAssign<&'a T>> Sub<Secret<T>> for T`), which is not allowed.

synedrion/src/uint/signed.rs
```diff
@@ -411,8 +381,8 @@ where
 {
     /// Returns a [`Signed`] with the same value, but twice the bit-width.
     /// Consumes `self`, but under the hood this method clones.
-    pub fn into_wide(self) -> Signed<T::Wide> {
-        let abs_result = self.abs().into_wide();
+    pub fn to_wide(self) -> Signed<T::Wide> {
```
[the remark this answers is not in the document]
fjarri: Strange that clippy didn't complain about it, because it did in other seemingly identical situations with other types

synedrion/src/tools/secret.rs
fjarri: The exponent will be put in a stack array when it's decomposed with a small radix for windowed exponentiation. So if it's secret, it's going to leave traces on the stack. With the multiplication there's less danger since `Uint`s implement `*Assign` traits which we use, so theoretically anything secret will be overwritten.

synedrion/src/tools/secret.rs
fjarri: Yeah, that's a good idea

synedrion/src/tools/secret.rs
fjarri: Hm, good catch. I think it is some remainder from the previous iterations, but I don't think it makes any sense now.

synedrion/src/tools/secret.rs
fjarri: We do take a reference here at least in some places where it comes from the caller, e.g. in `DecProof::new()`

synedrion/src/uint/signed.rs
dvdplm: I added the arithmetic ops (and `PartialOrd` and a few others) this summer; I used them in tests I think and also sort of thought it should be a "complete" number type, but I'm fine just implementing what we need too.

On a tangential note: do you think we will be able to get rid of our `Signed` impl at some point, given that `crypto-bigint` now supports it?

synedrion/src/uint/traits.rs
dvdplm: aha, ok, got it. Ty!

synedrion/src/paillier/encryption.rs
dvdplm: Of course, 🤦 .

But in that case, how about "new_bounded_with_randomizer"?

synedrion/src/tools/secret.rs
dvdplm: This is a somewhat subtle point, would be good to have your comment here in the code too for our future selves. Handling hazmat is *hard*!

synedrion/src/tools/secret.rs
fjarri: Fixed

synedrion/src/tools/secret.rs
fjarri: The explicit zeroization was removed

synedrion/src/tools/secret.rs
fjarri: Added

synedrion/src/uint/signed.rs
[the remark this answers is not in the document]
fjarri: Fixed

synedrion/src/paillier/encryption.rs
fjarri: That would imply that the result is Bounded. I am not a big fan of the name myself, but a grammatically correct variant "new_with_randomizer_and_bounded_secret" seems to be a little long?

synedrion/src/uint/signed.rs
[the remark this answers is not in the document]
fjarri: Actually no. The reason clippy doesn't complain is because `Signed` here implements `Copy`, so `to_*(self)` is appropriate. It implements `Copy` because of the `ConditionallySelectable` bound. So the name is fine as it is.

The problem is that it uses `abs()` inside which creates a copy on the stack that we don't really want, and `abs()` is used in a lot of methods which really shouldn't copy the number. I wonder if `Signed` can be rewritten to remove all this copying (and ideally the whole `ConditionallySelectable` bound).

synedrion/src/paillier/encryption.rs
dvdplm: Yep, you're right, that's too long. Let's keep the current name then.

synedrion/src/cggmp21/sigma/dec.rs
```diff
@@ -58,15 +62,18 @@ impl<P: SchemeParams> DecProof<P> {
...
+        let nu = Secret::init_with(|| Signed::random_bounded_bits_scaled(rng, P::L_BOUND + P::EPS_BOUND, hat_cap_n));
         let r = Randomizer::random(rng, pk0);
 
         let cap_s = setup.commit(y, &mu).to_wire();
         let cap_t = setup.commit(&alpha, &nu).to_wire();
-        let cap_a = Ciphertext::new_with_randomizer_signed(pk0, &alpha, &r.to_wire()).to_wire();
-        let gamma = P::scalar_from_signed(&alpha);
+        let cap_a = Ciphertext::new_with_randomizer_signed(pk0, &alpha, &r).to_wire();
+
+        // `alpha` is secret, but `gamma` only uncovers $\ell$ bits of `alpha`'s full $\ell + \eps$ bits,
```
dvdplm, line 74: So 256 bits of 3*256 bits? (`L_BOUND` = 256 and `EPS_BOUND` = 2*`L_BOUND`)?

dvdplm (approved):
Good stuff!

synedrion/src/cggmp21/sigma/dec.rs
fjarri: Yep. In any case, it's going in the proof which is supposed to be public, so I guess it shouldn't be a problem for it to be exposed.

synedrion/src/uint/signed.rs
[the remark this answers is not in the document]
fjarri: Filed https://github.com/entropyxyz/synedrion/issues/165 to not inflate this PR any more

synedrion/src/tools/secret.rs
fjarri: Added the comment

PR #164, opened 2024-12-06
Don't use slice indexing

### LINKED ISSUE #68: Check range errors at the protocol level
There are multiple places at the protocol level where we can't be sure about the self-consistency of party idx and the total number of parties. e.g.:
- the number of parties should be non-zero (maybe even >1)
- the party index must be within the number of parties
- the party index should not coincide with the current party's index (e.g. in `verify_direct_message()`/`verify_broadcast()`)
- the consistency of the party index/number of parties obtained from different sources (e.g. on creation of `presigning::Round1` we get those as separate arguments, and in a `KeyShare` object)
- the correct size and excluded index in the `HoleVec` objects passed to `finalize_to_result()`/`finalize_to_next_round()` methods

These should be asserted.

dvdplm:
Enable `clippy::indexing_slicing` lint in all non-test code. Also changes some methods to return `Result`.

Closes #68


synedrion/src/cggmp21/aux_gen.rs
```diff
@@ -626,6 +626,7 @@ impl<P: SchemeParams, I: PartyId + Serialize> Round<I> for Round3<P, I> {
     }
 }
 
+#[allow(clippy::indexing_slicing)]
```
fjarri, line 629: We can probably deal with it the same as with `unwrap_used`, `#![cfg_attr(not(test), warn(clippy::indexing_slicing))]` in `lib.rs`

synedrion/src/cggmp21/entities.rs
```diff
@@ -1,8 +1,10 @@
 use alloc::{
     collections::{BTreeMap, BTreeSet},
+    format,
     vec::Vec,
 };
 use core::{fmt::Debug, marker::PhantomData};
+use manul::session::LocalError;
```
fjarri, line 7: Not entirely sure if we should use our own error instead of reusing `manul`'s, but that'll work for now.

synedrion/src/cggmp21/entities.rs
```diff
@@ -125,23 +127,38 @@ pub(crate) struct PresigningValues<P: SchemeParams> {
...
+                change.public_share_changes.len()
+            )));
+        }
 
         let secret_share = self.secret_share + change.secret_share_change;
         let public_shares = self
             .public_shares
             .iter()
-            .map(|(id, public_share)| (id.clone(), *public_share + change.public_share_changes[id]))
+            .zip(change.public_share_changes)
```
fjarri, line 149: This assumes that `self.public_shares` and `change.public_share_changes` have the same set of keys, which we can't really be sure of

synedrion/src/cggmp21/interactive_signing.rs
```diff
@@ -365,6 +377,17 @@ struct Round2<P: SchemeParams, I: Ord> {
     all_cap_g: BTreeMap<I, Ciphertext<P::Paillier>>,
 }
 
+// TODO(dp): @reviewers convenient enough to include in `Round`?
+impl<P: SchemeParams, I: Ord + Debug> Round2<P, I> {
+    fn public_aux(&self, i: &I) -> Result<&PublicAuxInfoPrecomputed<P>, LocalError> {
```
fjarri, line 382: Can be a method of `Context` to avoid repetition?

synedrion/src/cggmp21/interactive_signing.rs
```diff
@@ -1009,6 +1090,21 @@ where
...
+        self.context
+            .aux_info
+            .public_aux
+            .get(i)
+            .ok_or_else(|| LocalError::new("Missing public_aux for party Id {i:?}"))
+    }
+
+    fn public_shares(&self, i: &I) -> Result<&Point, LocalError> {
+        self.context
```
fjarri, line 1102: I think you have the same code above where you could just call this method

synedrion/src/cggmp21/params.rs
```diff
@@ -136,8 +138,14 @@ pub(crate) fn uint_from_scalar<P: SchemeParams>(value: &Scalar) -> <P::Paillier
     let uint_len = repr.as_ref().len();
     let scalar_len = scalar_bytes.len();
 
-    debug_assert!(uint_len >= scalar_len);
-    repr.as_mut()[uint_len - scalar_len..].copy_from_slice(&scalar_bytes);
+    debug_assert!(
```
fjarri, line 141: Probably don't need this assertion if we already have an `expect()` just below.

synedrion/src/cggmp21/params.rs
```diff
@@ -263,6 +291,7 @@ pub(crate) fn secret_scalar_from_signed<P: SchemeParams>(
     // TODO: wrap in secrets properly
     let abs_value = scalar_from_uint::<P>(&value.expose_secret().abs());
     Secret::init_with(|| Scalar::conditional_select(&abs_value, &-abs_value, value.expose_secret().is_negative()))
+    // >>>>>>> master
```
fjarri, line 294: Merge artifact?

synedrion/src/cggmp21/params.rs
```diff
@@ -223,7 +246,12 @@ pub(crate) fn secret_uint_from_scalar<P: SchemeParams>(
     let scalar_len = scalar_bytes.expose_secret().len();
 
     debug_assert!(uint_len >= scalar_len);
-    repr.expose_secret_mut().as_mut()[uint_len - scalar_len..].copy_from_slice(scalar_bytes.expose_secret());
+    repr.expose_secret_mut()
+        .as_mut()
+        .get_mut(uint_len - scalar_len..)
+        // TODO(dp): @reviewers Do we need a better proof that this is true? If `Uint` is, say, 128 bits long…
```
fjarri, line 252: We do need some check for parameter consistency, but I'm not sure if it's possible to do in compile-time.

synedrion/src/curve/arithmetic.rs
```diff
@@ -114,6 +114,7 @@ impl Scalar {
         Secret::init_with(|| Self(*sk.as_nonzero_scalar().as_ref()))
     }
 
+    // TODO(dp): @reviewers Does this assume big endian order?
```
fjarri, line 117: I don't think so? It just assumes that it can deserialize whatever `to_bytes()` serialized

synedrion/src/tools/hashing.rs
```diff
@@ -155,25 +155,32 @@ where
...
+    // If the number of bits is not a multiple of 8, use a mask to zeroize the high bits in the
+    // gererated random bytestring, so that we don't have to reject too much.
     let mask = if n_bits & 7 != 0 {
         (1 << (n_bits & 7)) - 1
     } else {
         u8::MAX
     };
 
+    // TODO(dp): @reviewers this uses `to_le_bytes` (little-endian) but then the original code masks out bits
+    // from the *last* byte, so it'd mask out the *low* bits yeah? Changing to mask the first byte
```
fjarri, line 169: Little-endian means the last byte is the most significant one. (I know it's confusing, it doesn't mean "little in the end", but "little end is read first")

synedrion/src/tools/sss.rs
```diff
@@ -33,23 +34,25 @@ fn evaluate_polynomial<T>(coeffs: &[T], x: &Scalar) -> T
...
+    assert!(coeffs.len() > 1, "Expected coefficients to be non-empty");
     // Evaluate in reverse to save on multiplications.
     // Basically: a0 + a1 x + a2 x^2 + a3 x^3 == (((a3 x) + a2) x + a1) x + a0
-    let mut res = coeffs[coeffs.len() - 1];
-    for i in (0..(coeffs.len() - 1)).rev() {
-        res = res * x + coeffs[i];
-    }
-    res
+
+    let (acc, coeffs) = coeffs.split_last().expect("Coefficients is not empty");
```
fjarri, line 41: I probably wouldn't bother and just added an exception for clippy here :) But since you did, I guess we can keep it.

synedrion/src/uint/traits.rs
```diff
@@ -239,10 +238,11 @@ impl HasWide for U4096 {
     }
 }
 
-pub type U512Mod = MontyForm<{ nlimbs!(512) }>;
-pub type U1024Mod = MontyForm<{ nlimbs!(1024) }>;
-pub type U2048Mod = MontyForm<{ nlimbs!(2048) }>;
-pub type U4096Mod = MontyForm<{ nlimbs!(4096) }>;
+// TODO(dp): Suggest crypto-bigint update nlimbs! macro.
+pub type U512Mod = MontyForm<{ 512u32.div_ceil(Limb::BITS) as usize }>;
```
fjarri, line 242: This is not necessary, I pinned Rust version for the clippy CI action, it now runs on 1.81.

synedrion/src/cggmp21/entities.rs
dvdplm: Yeah I had the same thought: soon enough we'll need to have a think about which errors we want.

synedrion/src/cggmp21/interactive_signing.rs
dvdplm: That's on `Round2`, but if I move it to `Context` I could.

synedrion/src/cggmp21/params.rs
dvdplm: Worth investigating perhaps, like have a `const fn validate_config() -> bool` that checks for configuration parameter consistency, e.g. that `<P::Paillier as PaillierParams>::Uint` is bigger than a `Point`.

synedrion/src/curve/arithmetic.rs
dvdplm: `to_bytes` does `self.0.to_be_byte_array()` so it is indeed BE. I'll dig further in a separate PR.

synedrion/src/tools/hashing.rs
dvdplm: Right, hmm, so the high bits are in the last byte then and my `first_mut` is wrong. Changing to `last_mut`, good catch!

synedrion/src/uint/traits.rs
fjarri: I just run it with `+1.81`

Issue #162, opened 2024-11-27
Zeroize `MontyParams` of RSA primes


fjarri:
Currently blocked by https://github.com/RustCrypto/crypto-bigint/issues/704.

In the `paillier` module, we have some `MontyParams` structures with moduli `p` and `q`, and some precomputed constants containing these params. These need to be zeroized when `crypto-bigint` supports it.

AaronFeickert:
This functionality [has been merged](https://github.com/RustCrypto/crypto-bigint/pull/706).

PR #166, opened 2024-12-10
`Signed` improvements

### LINKED ISSUE #165: Zeroizing follow-up: `Signed`
#163 was getting big, so it had to be wrapped up, but there are still things to do with `Signed`.

`Signed` is used in `Secret<>` a lot, but its operations often create stack copies (mainly, whenever `conditional_select()` is involved). It would be great to remove the bound for `ConditionallySelectable` somehow. The selection is used in `abs()`, which is used in a lot of places - can we reduce that, or eliminate all of those instances?

Also, `checked_add` and `cheked_mul` seems to be overcomplicated? Since we're using 2-s complement, can we just `wrapping_add/mul` stuff and get a correct result, without explicitly dealing with signs?

fjarri:
- Split `Signed` and `Bounded` into secret and public types
- More careful handling of secret values in Signed (fixes #165)


synedrion/src/uint/public_signed.rs
```diff
@@ -0,0 +1,242 @@
...
+
+impl<T> Sub<PublicSigned<T>> for PublicSigned<T>
+where
+    T: Integer + Bounded,
+{
+    type Output = PublicSigned<T>;
+
+    fn sub(self, rhs: PublicSigned<T>) -> Self::Output {
+        self.checked_sub(&rhs)
+            .expect("the calling code ensured the bound is not overflown")
```
dvdplm, line 228: Hmm, can we do better than assuming the caller knows this? :/

synedrion/src/uint/secret_unsigned.rs
```diff
@@ -0,0 +1,114 @@
...
+use crypto_bigint::{subtle::Choice, Bounded, Integer, Monty, NonZero};
+use zeroize::Zeroize;
+
+use super::{HasWide, SecretSigned};
+use crate::tools::Secret;
+
+/// A bounded unsigned integer with sensitive data.
+#[derive(Debug, Clone)]
+pub(crate) struct SecretUnsigned<T: Zeroize> {
+    /// bound on the bit size of the value (that is, `value < 2^bound`)
```
dvdplm, line 12:
```suggestion
    // Bound on the bit size of the value (that is, `value < 2^bound`)
```

synedrion/src/uint/secret_unsigned.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// - The value of `T` is too big to be bounded by the provided bound.
+    pub fn new(value: Secret<T>, bound: u32) -> Option<Self> {
+        if bound > T::BITS || value.expose_secret().bits() > bound {
+            return None;
+        }
+        Some(Self { value, bound })
+    }
+
+    pub fn add_mod(&self, rhs: &Self, modulus: &Secret<NonZero<T>>) -> Self {
+        // Note: assuming that the bit size of the modulus is not secret
```
dvdplm, line 42: Move to doc-comment perhaps?

synedrion/src/uint/secret_unsigned.rs
```diff
@@ -0,0 +1,114 @@
...
+where
+    T: Zeroize + Integer + Bounded,
+{
+    type Output = Self;
+    fn bitand(self, rhs: T) -> Self::Output {
+        Self {
+            value: Secret::init_with(|| {
+                let mut result = self.value.expose_secret().clone();
+                result &= rhs;
+                result
```
dvdplm, line 73:
```suggestion
                self.value.expose_secret().clone() & rhs
```

synedrion/src/uint/traits.rs
```diff
@@ -16,152 +18,70 @@ pub trait ToMontgomery: Integer {
     }
 }
 
-/// Exponentiation functions for generic integers (in our case used for integers in Montgomery form
-/// with `Signed` exponents).
-pub trait Exponentiable<T>:
-    PowBoundedExp<T> + Invert<Output = CtOption<Self>> + ConditionallySelectable + Square + core::ops::Mul<Output = Self>
+pub trait Exponentiable<Exponent> {
```
dvdplm, line 21: Why delete the docs here and in the rest of the file? At least the panicking behaviour and CT-nature should be documented imo.

synedrion/src/paillier/encryption.rs
```diff
@@ -325,43 +304,49 @@ impl<P: PaillierParams> Ciphertext<P> {
...
         Self {
             pk: self.pk.clone(),
-            ciphertext: self.ciphertext.pow_signed(rhs),
+            ciphertext: self.ciphertext.pow(&rhs.to_wide()),
         }
     }
 
-    fn homomorphic_mul_unsigned(self, rhs: &Bounded<P::Uint>) -> Self {
-        let rhs_wide = rhs.to_wide();
+    pub fn homomorphic_mul_wide_public(&self, rhs: &PublicSigned<P::WideUint>) -> Self {
```
dvdplm, line 335: Missing docs

synedrion/src/tools/secret.rs
```diff
@@ -79,315 +73,222 @@ impl<T: Zeroize> Debug for Secret<T> {
...
 }
 
 impl<T: Zeroize + Integer<Monty: Zeroize>> Secret<T> {
     pub fn to_montgomery(&self, params: &<T::Monty as Monty>::Params) -> Secret<T::Monty> {
         // `self` has to be cloned and passed by value, which means it may be retained on the stack.
         // Can't help it with the current `Monty::new()` signature.
+
+        // NOTE: `params` is cloned here and can remain on the stack.
+        // See https://github.com/RustCrypto/crypto-bigint/issues/704;
+        // after that's fixed, we can properly treat secret `params`.
```
dvdplm, line 291: https://github.com/RustCrypto/crypto-bigint/issues/704 is fixed now; do we need to update this comment perhaps?

synedrion/src/uint/secret_signed.rs
```diff
@@ -0,0 +1,710 @@
...
+    WrappingMul, WrappingNeg, WrappingSub,
+};
+
+use super::{HasWide, PublicSigned, SecretUnsigned};
+use crate::tools::Secret;
+
+/// A wrapper over secret unsigned integers that treats two's complement numbers as negative.
+#[derive(Debug, Clone)]
+pub(crate) struct SecretSigned<T: Zeroize> {
+    /// bound on the bit size of the absolute value (that is, `abs(value) < 2^bound`)
```
dvdplm, line 17:
```suggestion
    // bound on the bit size of the absolute value (that is, `abs(value) < 2^bound`)
```

(or move this info to the docs)

synedrion/src/uint/secret_signed.rs
```diff
@@ -0,0 +1,710 @@
...
+    /// Panics if it is not the case.
+    ///
+    /// That is, the value must be within $[-2^{exp}, 2^{exp}]$
+    /// (See Section 2).
+    pub fn assert_exp_range(&self, exp: u32) {
+        let in_bound = self.in_bound(exp);
+        // Have to check for the ends of the range too
+        let is_end = self.abs().expose_secret().ct_eq(&(T::one() << exp));
+        assert!(bool::from(in_bound | is_end), "out of bounds $\\pm 2^{exp}$",)
+    }
```
dvdplm, line 165: I think that we should strive to avoid using asserts in non-test code, and instead prefer `Result` or `Option`. Doing this consistently is a lot of work and I am fine doing it bit by bit, but perhaps we can avoid adding *new* asserting code and instead return `Result` here?

synedrion/src/uint/secret_signed.rs
```diff
@@ -0,0 +1,710 @@
...
+
+    #[cfg(test)]
+    fn test_new_from_abs(abs_value: T, bound: u32, is_negative: bool) -> Option<Self> {
+        Self::new_from_abs(Secret::init_with(|| abs_value), bound, Choice::from(is_negative as u8))
+    }
+
+    #[cfg(test)]
+    fn test_new_from_unsigned(abs_value: T, bound: u32) -> Option<Self> {
+        Self::new_from_unsigned(Secret::init_with(|| abs_value), bound)
+    }
```
dvdplm, line 135: Could these be moved to the `tests` module you think?

dvdplm (commented):
Pheew, a big one!

Overall LGTM; left some nitpicks, suggestions and questions. 

There's some inconsistency in what is documented and what isn't, and I think we should avoid asserts in favour of `Result` wherever possible, although I don't consider it a blocker for this PR.

synedrion/src/cggmp21/conversion.rs
```diff
@@ -0,0 +1,133 @@
...
+pub(crate) fn scalar_from_wide_uint<P: SchemeParams>(value: &<P::Paillier as PaillierParams>::WideUint) -> Scalar {
+    let r = *value % P::CURVE_ORDER_WIDE;
+
+    let repr = r.to_be_bytes();
+    let uint_len = repr.as_ref().len();
+    let scalar_len = Scalar::repr_len();
+
+    // Can unwrap here since the value is within the Scalar range
+    Scalar::try_from_be_bytes(&repr.as_ref()[uint_len - scalar_len..])
+        .expect("the value was reduced modulo `CURVE_ORDER`, so it's a valid curve scalar")
```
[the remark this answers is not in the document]
fjarri: Technically, the same number, just stored in different types

synedrion/src/cggmp21/sigma/enc.rs
```diff
@@ -30,36 +29,36 @@ Public inputs:
...
-        k: &Secret<Signed<<P::Paillier as PaillierParams>::Uint>>,
+        k: &SecretSigned<<P::Paillier as PaillierParams>::Uint>,
         rho: &Randomizer<P::Paillier>,
         pk0: &PublicKeyPaillier<P::Paillier>,
         cap_k: &Ciphertext<P::Paillier>,
         setup: &RPParams<P::Paillier>,
         aux: &impl Hashable,
     ) -> Self {
-        k.expose_secret().assert_bound(P::L_BOUND);
+        k.assert_exp_range(P::L_BOUND);
```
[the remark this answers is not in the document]
fjarri: I renamed them because I didn't want confusion with the `bound` field of Signed. Maybe expand it to "exponent"? And "exp" is there to signify that it's the range of `2^x`, not `x`. Although maybe it's not that important since it's pretty much always that.

synedrion/src/paillier/rsa.rs
```diff
@@ -247,8 +241,10 @@ impl<P: PaillierParams> PublicModulus<P> {
         NonZero::new(self.modulus.0).expect("the modulus is non-zero")
     }
 
-    pub fn modulus_bounded(&self) -> Bounded<P::Uint> {
-        Bounded::new(self.modulus.0, P::MODULUS_BITS).expect("the modulus can be bounded by 2^MODULUS_BITS")
+    pub fn modulus_signed(&self) -> PublicSigned<P::WideUint> {
+        // Have to return WideUint, since Uint::BITS == P::MODULUS_BITS, so it won't fit in a Signed<Uint>.
+        PublicSigned::new_positive(self.modulus.0.to_wide(), P::MODULUS_BITS)
+            .expect("the modulus can be bounded by 2^MODULUS_BITS")
```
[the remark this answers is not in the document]
fjarri: Well, there is a bit of inefficiency with storage, but it's only used in exponentiation, so it doesn't really affect performance (because the bound is the same). I had a `PublicUnsigned` type for a while, but it was only used for this value alone, so I decided to get rid of it.

synedrion/src/uint/public_signed.rs
```diff
@@ -0,0 +1,242 @@
...
+        repr.as_mut()[(repr_len - bytes_len)..].copy_from_slice(&val.abs_bytes);
+        let abs_value = T::from_be_bytes(repr);
+
+        Self::new_from_abs(abs_value, val.bound, val.is_negative)
+            .ok_or_else(|| "Invalid values for the signed integer".into())
+    }
+}
+
+/// A wrapper over unsigned integers that treats two's complement numbers as negative.
+// In principle, Bounded could be separate from Signed, but we only use it internally,
```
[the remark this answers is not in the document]
fjarri: Yeah, copy-paste artifact (in both lines actually)

synedrion/src/uint/traits.rs
fjarri: Probably lost as I was gradually replacing the old trait, my bad

synedrion/src/uint/secret_signed.rs
fjarri: Yes, I thought of that too, we can make all the proof constructors (where this method is used) return `Result<..., LocalError>`, so they can fail.

synedrion/src/uint/secret_signed.rs
fjarri: Yes, I suppose

synedrion/src/uint/secret_signed.rs
dvdplm: Not sure how often this is called or how important it is for it to be CT, but I tinkered a bit with making it faster (made it return `bool` for convenience):

```rust
    pub fn assert_exp_range_fast(&self, exp: u32) -> bool {
        let abs = self.abs();
        let mask = T::one().wrapping_neg().wrapping_shl_vartime(exp);
        if abs.expose_secret() == &(mask.not() + T::one()) {
            return true;
        }
        let masked = abs & mask;
        masked.expose_secret() == &T::zero()
    }
```

Benchmark timings for `U2048`:

```
Bounds check/assert_exp_range, small exponent
                        time:   [420.31 ns 425.26 ns 431.98 ns]
                        change: [-1.3407% +0.3532% +2.1482%] (p = 0.72 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
Bounds check/assert_exp_range_fast, small exponent
                        time:   [266.69 ns 271.08 ns 277.42 ns]
                        change: [-0.9114% +0.2448% +1.4652%] (p = 0.71 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) high mild
  6 (6.00%) high severe
Bounds check/assert_exp_range, big exponent
                        time:   [713.33 ns 769.78 ns 821.11 ns]
                        change: [-8.6114% +0.4042% +10.744%] (p = 0.94 > 0.05)
                        No change in performance detected.
Bounds check/assert_exp_range_fast, big exponent
                        time:   [383.40 ns 421.51 ns 462.13 ns]
                        change: [-6.4458% +2.8628% +12.866%] (p = 0.55 > 0.05)
                        No change in performance detected.
```

synedrion/src/cggmp21/conversion.rs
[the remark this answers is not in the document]
fjarri: Changed to "curve order"

synedrion/src/cggmp21/conversion.rs
```diff
@@ -0,0 +1,133 @@
...
+pub(crate) fn secret_signed_from_scalar<P: SchemeParams>(
+    value: &Secret<Scalar>,
+) -> SecretSigned<<P::Paillier as PaillierParams>::Uint> {
+    SecretSigned::new_positive(secret_uint_from_scalar::<P>(value), ORDER.bits_vartime() as u32).expect(concat![
+        "a curve scalar value is smaller than the curve order, ",
+        "and the curve order fits in `PaillierParams::Uint`"
+    ])
+}
+
+pub(crate) fn secret_scalar_from_signed<P: SchemeParams>(
```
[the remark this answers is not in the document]
fjarri: This actually does not assume anything, the value gets reduced modulo curve order.

synedrion/src/paillier/rsa.rs
```diff
@@ -127,16 +127,12 @@ impl<P: PaillierParams> SecretPrimes<P> {
...
-        })
+    pub fn p_signed(&self) -> SecretSigned<P::Uint> {
+        SecretSigned::new_positive(self.p(), P::PRIME_BITS).expect("`P::PRIME_BITS` is valid")
     }
 
-    pub fn q_signed(&self) -> Secret<Signed<P::Uint>> {
-        Secret::init_with(|| {
-            Signed::new_positive(*self.q().expose_secret(), P::PRIME_BITS).expect("`P::PRIME_BITS` is valid")
-        })
+    pub fn q_signed(&self) -> SecretSigned<P::Uint> {
```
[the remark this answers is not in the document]
fjarri: I am not sure about this... Does this comment really add any information to what's already in the method signature?

synedrion/src/paillier/rsa.rs
```diff
@@ -127,16 +127,12 @@ impl<P: PaillierParams> SecretPrimes<P> {
         Secret::init_with(|| self.primes.q.expose_secret().to_wide())
     }
 
-    pub fn p_signed(&self) -> Secret<Signed<P::Uint>> {
-        Secret::init_with(|| {
-            Signed::new_positive(*self.p().expose_secret(), P::PRIME_BITS).expect("`P::PRIME_BITS` is valid")
-        })
+    pub fn p_signed(&self) -> SecretSigned<P::Uint> {
```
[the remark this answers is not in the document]
fjarri: Same question as above

synedrion/src/uint/public_signed.rs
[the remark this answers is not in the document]
fjarri: Fixed

synedrion/src/uint/public_signed.rs
fjarri: The other option is to use `checked_*` everywhere in the proofs/protocols, which will add a lot of code. 

The crucial thing here is that we know that the bounds are static and not data-dependent, so this will panic in tests 100% of the time. But I'm not sure how to articulate it properly.

This is true for `SecretSigned`, but may not be true for `PublicSigned` because they come from external sources. So maybe we can only enable checked ops for these. There's only a few instances where they are used, so perhaps it's the safest course of action.

synedrion/src/uint/secret_unsigned.rs
fjarri: Decided to just use `bits()` instead of `bits_vartime()` and avoid extra documentation

synedrion/src/uint/traits.rs
fjarri: Added back the trait docs.

synedrion/src/tools/secret.rs
fjarri: Removed the links, kept just the TODO for #162

synedrion/src/uint/secret_signed.rs
fjarri: Not sure what you mean, isn't it in the docs already? It's a doc comment

synedrion/src/uint/secret_signed.rs
fjarri: > Not sure how often this is called or how important it is for it to be CT

A bunch of times during the protocol, but on the order of dozens maybe. So given that it takes <1us, I don't think its performance is too important. It does handle secret data, so I think constant-timeness is appropriate.

synedrion/src/uint/secret_signed.rs
dvdplm: You are correct, my bad. Maybe capitalize to "Bound" though?

synedrion/src/uint/secret_signed.rs
dvdplm: Fair enough. Do you happen to know if `&` is CT?

synedrion/src/paillier/rsa.rs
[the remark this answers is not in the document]
dvdplm: No it doesn't, you're right. It looks a bit better in the rendered docs. One thing that might be worth mentioning though is a comment on why it's a signed value, or rather, the signed-ness is a technicality and not some funny "negative primes".

synedrion/src/cggmp21/sigma/enc.rs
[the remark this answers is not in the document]
dvdplm: I think "exponent" is better.

synedrion/src/cggmp21/conversion.rs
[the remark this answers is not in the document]
dvdplm: It assumes that the implementation is parametrized such that the curve order fits in a `Uint` yeah? I mean, lots of things would break if that wasn't the case so perhaps it's not worth pointing out?

synedrion/src/cggmp21/sigma/enc.rs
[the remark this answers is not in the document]
fjarri: Renamed

synedrion/src/uint/public_signed.rs
fjarri: Removed Sub and Mul impls for PublicSigned. They were only used once in `FacProof` creation, so there wasn't a danger of malicious input causing a panic, but we'll be safer this way.

synedrion/src/uint/secret_signed.rs
fjarri: I hope it is, or a lot of `crypto-bigint` stuff is not CT :) It's used in a lot of places

synedrion/src/paillier/encryption.rs
fjarri: Actually, the new `Exponentiable` makes it (and other specialized methods) redundant. Fixed.

synedrion/src/tools/secret.rs
```diff
@@ -116,6 +116,14 @@ where
 
 // Addition
 
+impl<T: Zeroize + Clone + for<'a> Add<&'a T, Output = T>> AddAssign<Secret<T>> for Secret<T> {
+    fn add_assign(&mut self, rhs: Secret<T>) {
+        // Can be done without reallocation when Integer is bound on AddAssign.
+        // See https://github.com/RustCrypto/crypto-bigint/pull/716
```
dvdplm, line 122: 👍

dvdplm (approved):
I think this is good to go yeah?

fjarri:
I was waiting for your response to the last round of changes, and I'm guessing you're fine with them. Merging.

PR #169, opened 2024-12-17
Argument grouping for ZK proofs

fjarri:
Stacked on top of #166 

Group secret and public inputs into structs to make it more transparent which variables are substituted at the points of their use (since the variable names in the protocols don't always correspond to the variable names in the proofs)


dvdplm (approved):
lgtm

PR #170, opened 2024-12-17
Updates for CGGMP'24

fjarri:
Fixes #157
Fixes #43
Fixes #91

- Updated paper references in `PAPER.md`, removed obsolete items.
- Updated to the new range definition (`±2^l` now means `[-2^(l-1)+1, 2^(l-1)]` instead of the previous `[-2^l, 2^l]`). This caused a bit of a chain reaction:
  - `SecretSigned::assert_exponent_range()` logic changed.
  - `SecretSigned::random_in_exp_range*()` logic changed, and also `exp` was changed to `exponent` to match the assertion name.
  - `PublicSigned::from_xof_reader_bounded()` changed its behavior to produce the number according to the new range definition, and was renamed to `from_xof_reader_in_range()`.
  - `PublicSigned::in_range_bits()` changed its behavior according to the new range definition, and was renamed to `is_in_exponent_range()`.
  - Wherever scalars are passed to proofs and `Ciphertext` constructor they are passed as `SecretSigned`, to comply with the range requirements in the proofs.
  - `conversions::secret_unsigned_from_scalar()` removed.
  - Added `SecretSigned::new_modulo()` constructor to make a signed number in range [-N/2, N/2] from an Uint in range [0, N).
  - Removed `Ciphertext::new()` and `decrypt()` (which took unsigned plaintexts), renamed `new_signed()` and `decrypt_signed()` to `new()` and `decrypt()`.
  - `Ciphertext::new_with_randomizer()` was renamed to `new_with_randomizer_unsigned()`, since it's now a special one, only used in P_mul. Renamed `new_with_randomizer_signed()` to `new_with_randomizer()`, and `new_public_with_randomizer_signed()` to `new_public_with_randomizer()`.
- Following that, wherever in the paper `e <-- ±q` is used, we are sampling the challenge as a `Scalar` using the new `Scalar::from_xof_reader()` method (and using that in `П^sch` as well).
- Updated paper references and notation in `aff-g` proof.
- Updated paper references in `prm` proof, and started using `BitVec` for its commitment.
- Updated `dec` proof (there were significant changes). Temporarily, it is located in `dec_new.rs`, will be moved to `dec.rs` when `Presigning` is updated.
- Added `elog` proof.
- Added `enc-elg` proof.
- Added `aff-g*` proof.
- Updated paper references and notation in `fac` proof, and implemented necessary changes to the algorithm (some variables are calculated differently, and the challenge is now a signed `Uint` and not a `Scalar`)
[...]

PR #173, opened 2025-01-17
Benchmarks for ZK proofs

dvdplm:
Adds a set of benchmarks for our ZK proofs, starting with `FacProof` and `AffG`. 

Also contains a synthetic benchmark that measures the performance of exponentiations with large exponents vs reducing such exponents by the totient.

Given the proof code is not part of the public API (and shouldn't be), this PR introduces a "private_benches" feature that allows a thin(-ish) criterion benchmark wrapper ("benches/zk_proofs.rs") to call code in `src/private_benches/mod.rs`. Here we have access to everything that is `pub(crate)`.

Run the benchmarks with `cargo bench --bench zk_proofs --features private_benches`.

The PR also initialize `tracing` during the benchmarks, which can be useful when doing optimization work.


fjarri:
Yes, unfortunately a feature is the only way to do it. There was one already, it was removed in https://github.com/entropyxyz/synedrion/pull/156 because all the relevant API became public.

synedrion/benches/zk_proofs.rs
```diff
@@ -0,0 +1,179 @@
...
+        bench_log_star,
+        bench_paillier_blum_modulus,
+        bench_mul_star,
+        bench_paillier_mul,
+        bench_prm,
+        bench_sch
+    );
+}
+
+// Running benchmarks without the test harness requires a main function at the top level, leading to
```
fjarri, line 173: I think you can avoid doing that by specifying `required-features = ["private-benches"]` for this benchmark in `Cargo.toml`. Then, if the feature is not enabled, the benchmark won't be compiled.

synedrion/Cargo.toml
```diff
@@ -50,9 +51,24 @@ k256 = { version = "0.13", default-features = false, features = ["ecdsa", "arith
...
 bench = true
 name = "bench"
 harness = false
 path = "benches/bench.rs"
+
+[[bench]]
+bench = true
+name = "pow"
+harness = false
+path = "benches/pow.rs"
```
fjarri, line 69: Why does this one use `path`, but the other doesn't?

synedrion/src/cggmp21/sigma/fac.rs
```diff
@@ -239,4 +240,29 @@ mod tests {
         let proof = FacProof::<Params>::new(&mut OsRng, &sk, &setup, &aux);
         assert!(proof.verify(pk, &setup, &aux));
     }
+
+    #[test_log::test]
+    fn prove_and_verify_wire_payload() {
```
fjarri, line 245: What was the reason for the test? Was there some serialization issue?

synedrion/benches/pow.rs
```diff
@@ -0,0 +1,125 @@
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
```
fjarri, line 1: I'm not sure if this benchmark needs to be in this repo. It worked as an investigation tool, but doesn't really have anything to do with the protocol itself.

synedrion/src/cggmp21/sigma/fac.rs
fjarri: Moreover, for the sake of consistency, if we have such a test in `fac`, we should have one for all the other ZK proofs as well.

synedrion/src/private_benches.rs
```diff
@@ -0,0 +1,758 @@
...
+            rng, x, y, rho, rho_y, pk0, pk1, cap_c, cap_d, cap_y, cap_x, rp_params, aux,
+        )
+    }
+
+    pub fn aff_g_proof_prove<R: CryptoRngCore + Clone + 'static>(rng: R) -> impl FnMut(&mut Bencher<'_>) {
+        move |b: &mut Bencher<'_>| {
+            b.iter_batched(
+                || {
+                    let rng2 = rng.clone();
+                    proof_inputs(rng2)
```
fjarri, line 127: I wonder if we can just create the inputs once - you won't need giant parameter lists, and the benchmarks will run faster. There's probably not much variation depending on the exact parameters, as long as the proof is correct?

synedrion/benches/pow.rs
dvdplm: Yeah I agree. I struggle to make up my mind on this. Like, yesterday I wrote a benchmark to check if multi-exponentiation would be faster when there is just one exponentiation (i.e. given `x * y^k`, is it faster to do  `x.mul(y.pow(k))` or `multiexp( (x, 1), (y, k) )`?). Totally useless as a long-term benchmark but useful to get an answer; so I just toss the code I wrote in the trash? I guess yes? Or possibly move to a PR against `crypto-bigint`?

synedrion/src/cggmp21/sigma/fac.rs
dvdplm: No, no particular serialization issues. I just realized that the code involved to deserialize and prepare data for verification is not wholly trivial and felt it'd be good to have a test for that part too.

Maybe I should convert all the tests to actually to the ser/deser? Or would you prefer duplicating the tests like I did here?

synedrion/Cargo.toml
dvdplm: Copy&paste. I don't think it's needed. Fixing.

synedrion/benches/zk_proofs.rs
dvdplm: It works! I spent waaaaay too much time trying to get that to work, tyvm! :)

synedrion/Cargo.toml
dvdplm: Actually I removed the `pow` bench. I'd like to keep it and similar "investigation-only" benches but need a think on where to put them.

synedrion/src/cggmp21/sigma/fac.rs
fjarri: Technically serialization/deserialization of proofs is already tested by running the protocols, but I see no harm in adding a ser/deser step in the current tests, as you proposed. No need do duplicate everything, I think.

synedrion/benches/pow.rs
dvdplm: Ended up removing this, at least for now.

synedrion/src/private_benches.rs
dvdplm: This seems like a clear win, and works well. But I noticed something on this particular benchmark: a 10% regression on verification. I guess it means that the proof verification is quite dependent on how the actual numbers play out and there's significant variability there. That suggests we should probably not optimise the setup code.

fjarri: I wonder why that would be the case. Sizes of resulting exponents probably shouldn't be that different, maybe it's the signs that account for it - the vartime operations during the verification can conditionally skip inversion.

Maybe [`iter_custom`](https://docs.rs/criterion/0.5.1/criterion/struct.Bencher.html#method.iter_custom) could work here?

dvdplm: The signs seem more likely than the sizes indeed. 

Bear in mind though that we are not setting the two MSBs in the primes yet so there's a little more variability in the modulus sizes than we'd like, but yeah, it's at most one bit more or less, so it's a bit odd.

Can you elaborate on why `iter_custom` would help here? I'm sure it'd *work*, just not sure I understand how it'd help clean up the code.

fjarri: You would be able to create the inputs and measure the execution time in the same closure, so you won't need `proof_inputs()` returning a tuple of ten things. I am not 100% sure it will work, but it might.

dvdplm: Hmm, I don't think it will help us here though. Or I'm still misunderstanding. It'd looks sort of like this:

```rust
             b.iter_custom(|iters| {
				 let (a, b, c, d, e, f, g, h, …, …) = proof_inputs(rng);
                 let start = Instant::now();
                 for _i in 0..iters {
                     black_box(AffGProof::<Params>::new(&mut rng, AffGSecretInputs{ /* some inputs here */}, AffGPublicInputs{ /* more inputs here */ }, &rp_params, &aux));
                 }
                 start.elapsed()
             })
```

Wouldn't this have the same problem that performance measurments become inaccurate when using the same inputs repeatedly?

(There's also a stupid problem because of how this benchmark is setup: I do not have access to `std::time::Instant` here; for some reason it's not in `core::time`.)

fjarri:
I guess this one goes in next now that #170 is merged?

synedrion/src/private_benches.rs
fjarri: I think what I had in mind is that you just disband `proof_inputs` and create the values inplace in the benchmark closure (since it's only used there). The inputs will still be randomly generated on the closure run, so no repetition.

dvdplm:
@fjarri Good to go?

PR #175, opened 2025-01-20
Don't use a PublicSigned for PrmProof

### LINKED ISSUE #174: PrmProof fails with production parameters
While working on benchmarks for `PrmProof` I ran into a failure: [this `expect` is not true](https://github.com/entropyxyz/synedrion/blob/master/synedrion/src/cggmp21/sigma/prm.rs#L106). What happens is that the `bound` is 2048 (because `PaillierProduction::PRIME_BITS` is 2048), but so is the totient. The preceding code is not wrong, the value of `p` is indeed smaller than the totient, it just quite often happens that the two primes play out to result in a 2048-bit number. In this scenario the `SecretUnsigned` cannot fit into a `PublicSigned` (no space for the sign bit).

We can:
- Widen the type so that a it fits in a `PublicSigned`
- Restrict ourselves to max 2047 bit totients (i.e. at least one of the primes must be 1023 bit I guess?)
- Re-examine the paper and proof and check if this value must be signed or if we can use a `PublicUnsigned` here

dvdplm:
Removes the `PublicSigned` wrapper from the `proof` member of `PrmProof`. When using production-sized parameters, the totient is often a 2048 bit number and so doesn't fit in a signed number type.

The Prm proof components are all positive and reduced modulo the totient, hence guaranteed to fit in a totient-sized number type (i.e. 2048 bits), so wrapping them is unnessecary.

Fixes #174.

### DIFF
--- a/synedrion/src/cggmp21/sigma/prm.rs
+++ b/synedrion/src/cggmp21/sigma/prm.rs
@@ -5,7 +5,7 @@
 
 use alloc::{vec, vec::Vec};
 
-use crypto_bigint::modular::Retrieve;
+use crypto_bigint::{modular::Retrieve, BitOps, PowBoundedExp};
 use digest::XofReader;
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


synedrion/src/cggmp21/sigma/prm.rs
```diff
@@ -124,9 +123,10 @@ impl<P: SchemeParams> PrmProof<P> {
 
         for ((e, z), a) in challenge.0.iter().zip(self.proof.iter()).zip(self.commitment.0.iter()) {
             let a = a.to_montgomery(monty_params);
-            let pwr = setup.base_randomizer().pow(z);
+            let pwr = setup.base_randomizer().pow_bounded_exp(z, z.bits_vartime());
             let test = if *e { pwr == a * setup.base_value() } else { pwr == a };
             if !test {
+                // TODO(dp): Should we make this CT perhaps?
```
fjarri, line 129: Proofs are public, so I don't think it is necessary

PR #176, opened 2025-01-21
Upgrade crypto-bigint to rc.8

dvdplm:
To be frank I'm not entirely sure where the `DefaultIsZeros` requirement comes from, which is annoying.

The rest is tiny and just fallout from the upgrade. The changes between rc6 and rc8 are bigger than I thought.


fjarri:
> To be frank I'm not entirely sure where the DefaultIsZeros requirement comes from, which is annoying.

Yeah it's strange. Usually it comes up if the type doesn't impl Zeroize, because there's a generic `impl<Z: DefaultIsZeros> Zeroize for Z`.

synedrion/src/paillier/params.rs
```diff
@@ -53,15 +51,14 @@ pub trait PaillierParams: core::fmt::Debug + PartialEq + Eq + Clone + Send + Syn
         + Encoding<Repr: Zeroize>
         + Hashable
         + HasWide<Wide = Self::WideUint>
-        // TODO: remove when https://github.com/RustCrypto/crypto-bigint/pull/709 is merged
-        + for<'a> RemAssign<&'a NonZero<Self::Uint>>
-        + InvMod
+        + InvMod<Output = Self::Uint>
```
dvdplm, line 54: `InvMod` changed since rc.6

dvdplm:
Closing in favour of #177 – I must have had a dirty Cargo.lock or some other devilry; re-doing it made the funny `DefaultIsZeroes` requirement go away.

PR #179, opened 2025-01-23
Prefer the Pow trait

dvdplm:
This PR removes the `Exponentiable` trait and its implementations for `SecretSigned`, `SecretUnsigned` and `PublicSigned`. It's up for debate whether this is actually better, but I think it's a little bit cleaner and may be helpful for tricks like "exponentiation with known totient"? The downside is that now the wrapper types are public.

### DIFF
--- a/synedrion/src/paillier/encryption.rs
+++ b/synedrion/src/paillier/encryption.rs
@@ -6,7 +6,7 @@ use core::{
 use crypto_bigint::{
     modular::Retrieve,
     subtle::{Choice, ConditionallyNegatable, ConstantTimeGreater},
-    Monty, ShrVartime,
+    Monty, Pow, PowBoundedExp, ShrVartime,
 };
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
@@ -17,7 +17,7 @@ use super::{
 };
 use crate::{
     tools::Secret,
-    uint::{Exponentiable, HasWide, PublicSigned, SecretSigned, SecretUnsigned, ToMontgomery},
+    uint::{HasWide, PublicSigned, SecretSigned, SecretUnsigned, ToMontgomery},
 };
 
 /// A public randomizer-like quantity used in ZK proofs.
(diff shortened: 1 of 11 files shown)


synedrion/src/paillier/params.rs
```diff
@@ -64,6 +64,13 @@ pub trait PaillierParams: core::fmt::Debug + PartialEq + Eq + Clone + Send + Syn
         + PowBoundedExp<Self::Uint>
         + PowBoundedExp<Self::WideUint>
         + PowBoundedExp<Self::ExtraWideUint>
+        + Pow<SecretUnsigned<Self::Uint>>
+        + Pow<PublicSigned<Self::Uint>>
+        + Pow<PublicSigned<Self::WideUint>>
+        + Pow<PublicSigned<Self::ExtraWideUint>>
+        + Pow<SecretSigned<Self::Uint>>
+        + Pow<SecretSigned<Self::WideUint>>
+        + Pow<SecretSigned<Self::ExtraWideUint>>
```
dvdplm, line 73: This is a bit nasty. :/

synedrion/src/paillier/encryption.rs
```diff
@@ -294,7 +294,9 @@ impl<P: PaillierParams> Ciphertext<P> {
         // To isolate `rho`, calculate `(rho^N)^(N^(-1)) mod N`.
         // The order of `Z_N` is `phi(N)`, so the inversion in the exponent is modulo `phi(N)`.
         let sk_inv_modulus = sk.inv_modulus();
-        let randomizer_mod = Secret::init_with(|| ciphertext_mod_n.pow(sk_inv_modulus));
+        let randomizer_mod = Secret::init_with(|| {
+            ciphertext_mod_n.pow_bounded_exp(sk_inv_modulus.expose_secret(), sk_inv_modulus.bound())
```
fjarri, line 298: Shouldn't `pow()` work here?

fjarri:
> and may be helpful for tricks like "exponentiation with known totient"?

Is it actually helpful? I think that's the crucial thing here. I would say that using macros and making internal types public is a downside, so what are we getting for it?

synedrion/src/paillier/encryption.rs
dvdplm: It should, good catch. This was a left-over from early experimentation.

dvdplm:
> Is it actually helpful? I think that's the crucial thing here. I would say that using macros and making internal types public is a downside, so what are we getting for it?

Indeed, I'm not convinced myself. I *think* it is useful (or even required) to have this to make using the `MultiExponentiate` feature possible/easier.

I'll put this back as draft and collect some more info on how/if it's useful.

dvdplm:
Closing this for now. Maybe some day we can do it in a less sucky way.

Issue #134, opened 2024-07-18
What is the best way to derive the initial chain code for BIP-32?


fjarri:
#132 introduced BIP-32 derivation for threshold key shares. As the initial chain code we use a hash of the full verifying key. Is this secure? Is there a better way to do it? 

One option is to use the `rid` value that is built as a kind of a shared secret during KeyInit. Would such double-purpose usage be secure?

PR #186, opened 2025-02-03
Generic EC

### LINKED ISSUE #27: Generalize the library to use any curve

dvdplm:
Make `synderion` generic over the elliptic curve type.

Very noisy PR, but the gist of it is that we make the `SchemeParams` trait agnostic over the elliptic curve used and thread that change across the code base while removing all explicit dependencies on `k256`. 
Introduces a new generic parameter `P: SchemeParams` in many places, hence most of the noise. 

The main areas to consider during review are:

- Deserialization bounds. I have to admit that I often have very little clue as to why certain `#[serde(bound(deserialize(STUFF))]` are necessary or even why they work beyond "serde magic". Not great.
- bip32 related functionality. BIP32 is only really well-defined for secp256k1 and while the code in this PR works for other curves as well, nothing but secp256k1 should ever be used in production without careful analysis.
- The added `HashOutput` associated type on `SchemeParams` should be considered temporary. The goal is to remove the need for it and then we can remove it entirely.
- The added `WideCurveUint` associated type on `SchemeParams` is there to allow reduced bias when generating scalars from extendable hashes. It's not great to have to do this, but I have found no better way. :/
- There is an unfortunate delay in releasing updates to the RustCrypto stack following the release of crypto-bigint 0.6, which means that in this PR we're juggling two different versions of the bigint library. Hopefully we can get rid of this Real Soon™.

One outstanding question not addressed here is how to tweak the `TestParams` for optimal speed. For `TinyCurve64` these settings seem to work:

```
-    const L_BOUND: u32 = 256;
-    const LP_BOUND: u32 = 256;
-    const EPS_BOUND: u32 = 320;
+    const L_BOUND: u32 = 64;
+    const LP_BOUND: u32 = 320;
+    const EPS_BOUND: u32 = 128;
```
…along with a `PRIME_BITS` setting of `256`. 

Closes #27.


synedrion/src/cggmp21/entities.rs
```diff
@@ -21,18 +21,26 @@ use crate::{
...
-    shares: BTreeMap<I, Point>, // `X_j`
-    // TODO (#27): this won't be needed when Scalar/Point are a part of `P`
-    phantom: PhantomData<P>,
+#[serde(bound(deserialize = "I: for<'x> Deserialize<'x>"))]
+pub struct PublicKeyShares<P, I>
+where
+    P: SchemeParams,
+    I: Ord,
+{
+    shares: BTreeMap<I, Point<P>>, // `X_j`
```
dvdplm, line 43: Maybe at this point this should just be a wrapper over `BTreeMap`, `pub struct PublicKeyShares<P, I>(BTreeMap<I, Point<P>>)`?

synedrion/benches/bench.rs
```diff
@@ -5,13 +5,15 @@ use manul::{
     dev::{run_sync, BinaryFormat, TestSessionParams, TestSigner, TestVerifier},
     signature::Keypair,
 };
-use rand_core::OsRng;
-use synedrion::{AuxGen, AuxInfo, InteractiveSigning, KeyInit, KeyShare, TestParams};
+use primeorder::FieldBytes;
```
fjarri, line 8: We only need `primeorder` for that, and it's just a type alias, perhaps we can just define it ourselves.

synedrion/src/curve/arithmetic.rs
```diff
@@ -1,132 +1,169 @@
 use alloc::{format, string::String, vec, vec::Vec};
-use core::{
-    default::Default,
-    ops::{Add, Mul, Neg, Sub},
-};
+use core::ops::{Add, Mul, Neg, Rem, Sub};
+#[cfg(test)]
```
fjarri, line 3: That'll only apply to the first `use` statement.

synedrion/src/curve/arithmetic.rs
```diff
@@ -1,132 +1,169 @@
...
 
-pub(crate) type Curve = Secp256k1;
-pub(crate) type BackendScalar = k256::Scalar;
-pub(crate) type BackendPoint = k256::ProjectivePoint;
-pub(crate) type CompressedPointSize = <FieldBytesSize<Secp256k1> as ModulusSize>::CompressedPointSize;
-
-pub(crate) const ORDER: U256 = Secp256k1::ORDER;
-
-impl HashableType for Curve {
+impl HashableType for k256::Secp256k1 {
```
fjarri, line 35: Can it be defined generically for any `Curve` implementor?

synedrion/src/curve/arithmetic.rs
```diff
@@ -1,132 +1,169 @@
...
+    /// modulo the curve order to ensure a valid, unbiased scalar.
     pub fn from_xof_reader(reader: &mut impl XofReader) -> Self {
-        let mut bytes = k256::WideBytes::default();
-        reader.read(&mut bytes);
-        Self(<BackendScalar as Reduce<U512>>::reduce_bytes(&bytes))
+        let bytes_lo = reader.read_boxed(Self::repr_len());
+        let bytes_lo = GenericArray::<_, FieldBytesSize<P::Curve>>::from_slice(&bytes_lo);
+        let bytes_hi = reader.read_boxed(Self::repr_len());
+        let bytes_hi = GenericArray::<_, FieldBytesSize<P::Curve>>::from_slice(&bytes_hi);
+        let uint_lo = <P::Curve as Curve>::Uint::decode_field_bytes(bytes_lo);
```
fjarri, line 116: Does that work with non-reduced data?

synedrion/src/curve/arithmetic.rs
```diff
@@ -1,132 +1,169 @@
...
+        let uint_hi = <P::Curve as Curve>::Uint::decode_field_bytes(bytes_hi);
+        let wide_uint = uint_lo.concat(&uint_hi);
+        // TODO: When the elliptic curve stack upgrades to crypto-bigint v0.6 we can use RemMixed and
```
fjarri, line 119: You can use `crypto-bigint 0.6` (the one we use for Paillier, for example) all the way and in the last line convert it to the old `Uint`, like in `params::convert_uint()`

synedrion/src/curve/arithmetic.rs
```diff
@@ -410,4 +546,31 @@ mod test {
         let s_from_le_bytes = Scalar::try_from_be_bytes(&le_bytes).expect("bytes are valid-ish");
         assert_ne!(s, s_from_le_bytes, "Using LE bytes should not work")
     }
+
+    #[test]
+    fn to_and_from_bytes_tiny32() {
```
fjarri, line 551: Looks like it could be just a generic function called with different `SchemeParams` types. 

Also not sure what this test is actually testing in the second part. Why all this business with inverse? It can actually create an invalid Scalar representation, if it reverses something like `01 FF FF ... FF`.

synedrion/src/lib.rs
```diff
@@ -34,9 +32,11 @@ pub use cggmp21::{
     KeyRefreshProtocol, KeyShare, KeyShareChange, PrehashedMessage, ProductionParams112, SchemeParams, TestParams,
 };
 pub use curve::RecoverableSignature;
-pub use www02::{DeriveChildKey, KeyResharing, KeyResharingProtocol, NewHolder, OldHolder, ThresholdKeyShare};
+pub use www02::DeriveChildKey;
```
fjarri, line 35: Can be merged with the next line

synedrion/src/tools/secret.rs
```diff
@@ -25,7 +26,10 @@ use crate::{
 /// - Safe serialization/deserialization (down to `serde` API; what happens there we cannot control)
 pub(crate) struct Secret<T: Zeroize>(SecretBox<T>);
 
-impl<T: Zeroize> Secret<T> {
+impl<T> Secret<T>
```
fjarri, line 29: Btw, do you use some tool to move the type bounds into the `where` clause? I would like to make it consistent throughout the codebase, but I couldn't find a lint in `clippy` or `rustfmt` setting that would do that.

synedrion/src/tools/sss.rs
```diff
@@ -83,31 +99,42 @@ impl Polynomial {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub(crate) struct PublicPolynomial(Vec<Point>);
+#[serde(bound(deserialize = "for<'x> P: Deserialize<'x>"))]
```
fjarri, line 102: Strange that you need an explicit bound for `Deserialize` but not for `Serialize`

synedrion/src/tools/sss.rs
```diff
@@ -83,31 +99,42 @@ impl Polynomial {
...
+{
     let polynomial = Polynomial::random(rng, secret, threshold);
     indices.iter().map(|idx| (*idx, polynomial.evaluate(idx))).collect()
 }
 
-pub(crate) fn interpolation_coeff(share_ids: &BTreeSet<ShareId>, share_id: &ShareId) -> Scalar {
+pub(crate) fn interpolation_coeff<P>(share_ids: &BTreeSet<ShareId<P>>, share_id: &ShareId<P>) -> Scalar<P>
+where
+    P: SchemeParams,
+    <P::Curve as CurveArithmetic>::Scalar: Copy,
```
fjarri, line 136: Do we really need this bound? Can we just clone?

synedrion/src/uint/traits.rs
```diff
@@ -175,7 +175,6 @@ impl HasWide for U4096 {
     type Wide = U8192;
 }
 
-// TODO(dp): Suggest crypto-bigint update nlimbs! macro.
```
fjarri, line 178: We can probably bring back `nlimbs!` usage, it doesn't produce clippy warnings anymore.

synedrion/src/www02.rs
```diff
@@ -1,5 +1,8 @@
 mod entities;
 pub(crate) mod key_resharing;
 
-pub use entities::{DeriveChildKey, ThresholdKeyShare};
+pub use bip32::DeriveChildKey;
+pub use entities::ThresholdKeyShare;
 pub use key_resharing::{KeyResharing, KeyResharingProtocol, NewHolder, OldHolder};
+
+mod bip32;
```
fjarri, line 8: Not sure it should be a part of `www02` - look more suitable for `curve` submodule.

synedrion/src/www02/bip32.rs
```diff
@@ -0,0 +1,206 @@
...
+    derivation_path: &DerivationPath,
+) -> Result<Vec<PrivateKeyBytes>, bip32::Error>
+where
+    C: CurveArithmetic + PrimeCurve + DigestPrimitive,
+    <C as CurveArithmetic>::AffinePoint: FromEncodedPoint<C> + ToEncodedPoint<C>,
+    <C as Curve>::FieldBytesSize: ModulusSize,
+{
+    let mut public_key = public_key.clone();
+
+    // Note: deriving the initial chain code from public information. Is this okay? <–– PROBABLY NOT OK?
```
fjarri, line 165: Should be marked as `// TODO (#134):`

synedrion/src/www02/bip32.rs
```diff
@@ -0,0 +1,206 @@
...
+where
+    C: CurveArithmetic + PrimeCurve,
+    <C as Curve>::FieldBytesSize: ModulusSize,
+    <C as CurveArithmetic>::AffinePoint: FromEncodedPoint<C> + ToEncodedPoint<C>,
+{
+    let mut public_key = public_key;
+    for tweak in tweaks {
+        public_key = public_key.derive_child(*tweak)?;
+    }
+    let offset = bip32::KEY_SIZE - <C as Curve>::FieldBytesSize::USIZE;
```
fjarri, line 202: I'm not sure I understand how this works. `bip32::KEY_SIZE = 32`, `<C as Curve>::FieldBytesSize::USIZE` is 32 for `k256` or 24 for `tiny-curve`. Then the length of `public_key.to_bytes()` is 33, with all of these bytes being meaningful for `k256`. Then we just drop the first byte (the sign of `y`) and it works as a SEC1 representation?

synedrion/src/www02/bip32.rs
```diff
@@ -0,0 +1,206 @@
...
+    use super::*;
+    pub trait Sealed {}
+    impl Sealed for VerifyingKey<tiny_curve::TinyCurve64> {}
+    impl Sealed for VerifyingKey<k256::Secp256k1> {}
+    impl Sealed for SigningKey<tiny_curve::TinyCurve64> {}
+    impl Sealed for SigningKey<k256::Secp256k1> {}
+}
+
+/// Trait for types that can derive BIP32 style "tweaks" from public keys.
+pub trait PubTweakable: sealed::Sealed {
```
fjarri, line 39: Maybe "Public", for symmetry with the "Secret" counterpart?

synedrion/src/www02/bip32.rs
```diff
@@ -0,0 +1,206 @@
...
+        let tweaks = derive_tweaks::<P::Curve>(&tweakable_pk, derivation_path)?;
+
+        // Will fail here if secret share is zero
+        let secret_share = self.secret_share.clone().to_signing_key().ok_or(bip32::Error::Crypto)?;
+        let mut tweakable_sk = secret_share.tweakable_sk();
+        for tweak in &tweaks {
+            tweakable_sk = tweakable_sk.derive_child(*tweak)?;
+        }
+        let bytes = tweakable_sk.to_bytes();
+        let bytes = bytes.get(32 - Scalar::<P>::repr_len()..).ok_or(bip32::Error::Decode)?;
```
fjarri, line 98: Wouldn't it be easier to just unwrap `SecretTweakable` into `SecretKey`, via another method in the `SecretTweakable` trait?

synedrion/src/www02/bip32.rs
fjarri: Also the question above about unwrapping applies here as well - I don't think you need to de/serialize at all.

synedrion/tests/threshold.rs
```diff
@@ -179,7 +190,9 @@ fn full_sequence() {
         })
         .collect();
 
-    info!("\nRunning InteractiveSigning\n");
+    info!("AuxGen took {:?}", now.elapsed()); // ~42 sec
```
fjarri, line 193: 42s applies to ProductionParams, not TestParams, I would hope?

synedrion/src/www02/key_resharing.rs
```diff
@@ -321,14 +330,15 @@ impl<P: SchemeParams, I: PartyId> Round<I> for Round1<P, I> {
...
-            let subshare = old_holder.polynomial.evaluate(their_share_id);
+            let subshare: Secret<Scalar<P>> = old_holder.polynomial.evaluate(their_share_id);
             let dm = DirectMessage::new(serializer, Round1DirectMessage { subshare })?;
             Ok((dm, None))
         } else {
             Ok((DirectMessage::none(), None))
         }
     }
 
+    #[allow(unused)]
```
fjarri, line 341: What's that for?

synedrion/src/cggmp21/params.rs
```diff
@@ -104,10 +114,39 @@ impl PaillierParams for PaillierProduction112 {
...
 
 /// Signing scheme parameters.
-// TODO (#27): this trait can include curve scalar/point types as well,
-// but for now they are hardcoded to `k256`.
-pub trait SchemeParams: Debug + Clone + Send + PartialEq + Eq + Send + Sync + 'static {
-    /// Bits of security the parameters ensure.
+pub trait SchemeParams:
+    'static + Debug + Clone + Send + PartialEq + Eq + Send + Sync + Ord + Copy + Serialize + for<'x> Deserialize<'x>
+where
+    <Self::Curve as CurveArithmetic>::ProjectivePoint: FromEncodedPoint<Self::Curve>,
```
fjarri, line 120: Some of the bounds look like they could be removed. The `Copy` ones, for example.

synedrion/src/cggmp21/params.rs
```diff
@@ -164,45 +203,80 @@ pub struct TestParams;
 // - Range checks will fail with the probability $q / 2^\eps$, so $\eps$ should be large enough.
 // - P^{fac} assumes $N ~ 2^{4 \ell + 2 \eps}$
 impl SchemeParams for TestParams {
+    type Curve = TinyCurve64;
+    type WideCurveUint = bigintv05::U384;
+    // TODO: 8*24 = 192, this is to work around an issue with the ModulusSize-trait. This should be ideally be 8 bytes long.
+    type HashOutput = [u8; 24];
     const SECURITY_BITS: usize = 16;
     const SECURITY_PARAMETER: usize = 10;
     const L_BOUND: u32 = 256;
```
fjarri, line 212: These bounds can be amended now that we have a small-order curve. l = log2(q) (so 32 or 64), lp = l * 5, eps = l * 2

fjarri: And the `PaillierTest` uint sizes adjusted accordingly.

synedrion/src/cggmp21/sigma/aff_g.rs
```diff
@@ -46,15 +46,16 @@ pub(crate) struct AffGPublicInputs<'a, P: SchemeParams> {
     /// Paillier ciphertext $Y = enc_1(y, \rho_y)$.
     pub cap_y: &'a Ciphertext<P::Paillier>,
     /// Point $X = g * x$, where $g$ is the curve generator.
-    pub cap_x: &'a Point,
+    pub cap_x: &'a Point<P>,
 }
 
 /// ZK proof: Paillier Affine Operation with Group Commitment in Range.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(bound(deserialize = "for<'x> P: Deserialize<'x>"))]
```
fjarri, line 54: It's kind of weird to see these bounds everywhere. I thought about dealing with #10 by bounding `SchemeParams` on `Serialize`/`Deserialize`, which you did, but I would expect that to eliminate the need for the explicit bounds?

synedrion/src/cggmp21.rs
```diff
@@ -36,6 +36,8 @@ pub use interactive_signing::{
 };
 pub use key_init::{KeyInit, KeyInitAssociatedData, KeyInitProtocol};
 pub use key_refresh::{KeyRefresh, KeyRefreshAssociatedData, KeyRefreshProtocol};
+#[cfg(test)]
+pub(crate) use params::TestParams32;
```
fjarri, line 40: Should have the same visibility as `TestParams` (which should perhaps be renamed to `TestParams64`?). `TestParams` usually would be something one would gate behind a `dev` feature, but it's just a single type, so I didn't bother creating a feature at the time.

synedrion/src/cggmp21/entities.rs
fjarri: Yeah, probably.

synedrion/src/www02/bip32.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use super::ThresholdKeyShare;
+
+/// Used for deriving child keys from a parent type.
+pub trait DeriveChildKey<C: CurveArithmetic + PrimeCurve>: Sized {
+    /// Return a verifying key derived from the given type using the BIP-32 scheme.
+    fn derive_verifying_key_bip32(&self, derivation_path: &DerivationPath) -> Result<VerifyingKey<C>, bip32::Error>;
+}
+
+mod sealed {
```
fjarri, line 29: What's up with sealing? Do these traits show up in the docs otherwise?

fjarri:
> One outstanding question not addressed here is how to tweak the TestParams for optimal speed. For TinyCurve64 these settings seem to work:

Yes, that looks correct, I added a comment in `params.rs` before reading the PR description :)

> …along with a PRIME_BITS setting of 256.

Why not 64?

Also, there are three remaining comments mentioning `#27` (in `curve.rs`, `sigma/fac.rs`, and `paillier/encryption.rs`) which can be now removed/amended.

synedrion/src/cggmp21.rs
dvdplm: In my head we'd put `k256` behind a feature, then add a default feature and include `k256`. Then we'd pick one of the tiny-curves that we're happy with and always have that active in test mode. 

Adding `TestParams32` was more of a check that things work well with `TinyCurve32` (also so we don't accidentally build in a dependency on `TinyCurve64`).

I thought `TestParams` was used in tests outside `synedrion` too, in entropy-core?

Here's what I suggest we do:
- remove `TestParams32` from this PR
- leave `TestParams` (both the name and the pub export)
- add an issue to figure out the best "strategy" for the features we want to introduce in v0.3

synedrion/src/cggmp21/params.rs
dvdplm: I assume you mean `Copy` on `SchemeParams`? Yeah I was on the fence on that, and tried both with and without; in the end I think that the reduced noise all over the code base from having it be `Copy` is worth it.

synedrion/src/cggmp21/params.rs
dvdplm: Finally indeed! I played with it a little bit last night but ended up stashing the changes because it just made the review even bigger. Follow up PR yes?

synedrion/src/cggmp21/sigma/aff_g.rs
dvdplm: This stuff was driving me nuts. I cannot tell you why this works or if there are other/better ways of doing it. It was by far the most frustrating thing about this work. You add a bound in one place and it typechecks, but then breaks somewhere else, you fix it over there in a way that makes sense and now it breaks somewhere else...

synedrion/src/curve/arithmetic.rs
dvdplm: Yes, that's on purpose, so that `TestParams` can be exported.

synedrion/src/curve/arithmetic.rs
dvdplm: I think problem here was that there's a blanket impl that conflicts? But maybe I'm wrong. I'll try!

synedrion/src/curve/arithmetic.rs
dvdplm: I think this test should be removed. It's just a copy of the existing test above + tweaks and a different parametrization.

synedrion/src/lib.rs
dvdplm: :+1

synedrion/src/tools/secret.rs
dvdplm: Yes, but it only works as a one-off thing, mapped to a key combo I have to press for each spot in the code. Having a way to do it across the code base would be real nice. Like, when I write code it's often convenient to type the bounds directly, but when reading code it's much easier (for me) to scan for the `where` clause; so automation would work well there for me.

synedrion/src/tools/sss.rs
dvdplm: Agreed. See other comment.

synedrion/src/tools/sss.rs
dvdplm: We don't, good catch!

synedrion/src/www02.rs
dvdplm: Agreed. Moving.

synedrion/src/www02/entities.rs
```diff
@@ -24,21 +21,24 @@ use crate::{
...
+    pub(crate) secret_share: Secret<Scalar<P>>,
+    pub(crate) share_ids: BTreeMap<I, ShareId<P>>,
+    pub(crate) public_shares: BTreeMap<I, Point<P>>,
 }
 
-impl<P: SchemeParams, I: Clone + Ord + PartialEq + Debug> ThresholdKeyShare<P, I> {
+impl<P, I> ThresholdKeyShare<P, I>
+where
+    P: SchemeParams,
+    I: Clone + Ord + PartialEq + Debug + Serialize + for<'x> Deserialize<'x>,
```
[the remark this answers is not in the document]
dvdplm: Nice. `PartyId` is a bit more restrictive though. Still, cleaner.

synedrion/tests/threshold.rs
dvdplm: Yes! Sorry, I'll remove them. But yeah, it's slow af.

synedrion/src/cggmp21/params.rs
dvdplm: Added https://github.com/entropyxyz/synedrion/issues/192

synedrion/src/www02/bip32.rs
dvdplm: The idea was to lock down the types for which it's ok to impl them, not really about hiding from docs. You think it's overkill? They have to be pub because the types they are impl'd for are pub.

synedrion/src/www02/bip32.rs
dvdplm: Let me spell out my reasoning and you can shoot it down!

- For `k256` the offset is `0`, so `bytes.get(offset..)` will be the whole slice of 33 bytes yes? Should behave the same as the old code.
- For `TinyCurve64` the offset is `8`, so in this case we truncate the high 8 bytes. This is likely not ok for a proper SEC1 repr, but the premise here is that BIP32 isn't really defined for anything other than secp256k1 so it's unclear what "valid" would even mean.

synedrion/src/www02/bip32.rs
dvdplm: Can you elaborate a bit on this (and "I don't think you need to de/serialize at all" above), not sure how you mean!

synedrion/src/cggmp21/params.rs
fjarri: No, the `Copy` ones in the `where` clause. The `Copy` on `SchemeParams` is reasonable, I think clippy will complain if it's not present.

synedrion/src/www02/bip32.rs
fjarri: See https://github.com/fjarri/synedrion/commit/fa153228a3231c81a66baefaace6e44fe713f7a1 (naming is rough, bounds may be adjusted, and perhaps some arguments can be taken by value). Basically, you have a trait to convert `VerifyingKey<C>` to something `bip32` accepts, so I just add another method to convert that something back to `VerifyingKey<C>`, which we can do either by just returning the same object (for `k256`) or unwrapping the newtype (for `tiny-curve`). Same for `SigningKey`.

synedrion/src/www02/bip32.rs
fjarri: Hm, perhaps it does work, but in any case it's not necessary - we can avoid serialization entirely, see the other comment.

synedrion/src/cggmp21/sigma/aff_g.rs
fjarri: Filed https://github.com/serde-rs/serde/issues/2899. These bounds can be avoided if we don't bound `SchemeParams` on `Deserialize`, but then other explicit bounds will be needed - but we already had those, so maybe that's the best course of action.

synedrion/src/cggmp21/params.rs
dvdplm: Removed, good catch.

synedrion/src/cggmp21/sigma/aff_g.rs
dvdplm: Did this in [https://github.com/entropyxyz/synedrion/pull/186/commits/7a1a5745e65fc7f7c78308a779450b26aaccd1d9] and it's probably a bit better than having a `Deserialize` on `SchemeParams` – at least it's a bit more explicit. 
That said it's both noisy and weird: sometimes explicit bounds are required and [sometimes not](https://github.com/entropyxyz/synedrion/pull/186/commits/7a1a5745e65fc7f7c78308a779450b26aaccd1d9#diff-f55b0977d14a290d03f0fcbded011e7fceb653ba15d2b4693d059183405de423R61), and I don't quite understand why.

synedrion/src/www02/bip32.rs
dvdplm: Cherry picked! ty

synedrion/src/curve/arithmetic.rs
dvdplm: I think it's probably fine, but better be prudent. I refactored using `from_be_byte_array` instead.

Also found a bug: forgot that these uints are v0.5, where the order is hi+lo as opposed to v0.6 where it's lo+hi.

synedrion/benches/bench.rs
dvdplm: You mean like `GenericArray<u8, FieldBytesSize<C>>`? What do we gain from that? We depend on `primeorder` anyway yeah?

synedrion/Cargo.toml
```diff
@@ -26,14 +25,19 @@ tracing = { version = "0.1.41", default-features = false }
 rand_chacha = { version = "0.3", default-features = false }
 rand = { version = "0.8", default-features = false }
 
+tiny-curve = { version = "0.1", features = [
```
fjarri, line 28: I published v0.2 with all the recent updates

synedrion/src/curve/arithmetic.rs
```diff
@@ -1,132 +1,169 @@
...
     /// Convert a 32-byte hash digest into a scalar as per SEC1:
     /// <https://www.secg.org/sec1-v2.pdf< Section 4.1.3 steps 5-6 page 45
     ///
-    /// SEC1 specifies to subtract the secp256k1 modulus when the byte array
+    /// SEC1 specifies to subtract the curve modulus when the byte array
     /// is larger than the modulus.
-    pub fn from_reduced_bytes(bytes: &[u8; 32]) -> Self {
-        let arr = GenericArray::<u8, FieldBytesSize<Secp256k1>>::from(*bytes);
-        Self(<BackendScalar as Reduce<U256>>::reduce_bytes(&arr))
+    pub fn from_reduced_bytes(bytes: impl AsRef<[u8]>) -> Self {
```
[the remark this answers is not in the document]
fjarri: Yeah, we'll leave it for later.

PR #191, opened 2025-02-14
Simplify ZK proof benches

fjarri:
Flatten out the code and reduce repetitions in ZK proof benches, and also keep the `criterion` dependency in `dev-dependencies` only. Using the `iter_custom()` approach for now since `iter_batched()` requires either returning closures, or introducing additional structures holding prepared parameters for each ZK proof which seems more complicated. 

Note: I pre-calculate two Paillier keys and two RP params and save them in a `LazyLock` (since that's generally the slowest part of the setup). This means that tests lose a bit of variation, but I don't expect much variation on the Paillier key, since that would mean timing attacks are possible.

It would be nice to somehow merge the code in these benchmarks with the test code in ZK proof modules, but that is left for a follow-up PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1159,12 +1159,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "smallvec"
-version = "1.13.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
-
-[[package]]
 name = "spki"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1223,7 +1217,6 @@ dependencies = [
  "tiny-curve",
  "tokio",
  "tracing",
- "tracing-subscriber",
  "zeroize",
 ]
 
@@ -1338,18 +1331,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e672c95779cf947c5311f83787af4fa8fffd12fb27e4993211a84bdfd9610f9c"
 dependencies = [
  "once_cell",
- "valuable",
-]
-
-[[package]]
-name = "tracing-log"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee855f1f400bd0e5c02d150ae5de3840039a3f54b025156404e34c23c03f47c3"
-dependencies = [
- "log",
- "once_cell",
- "tracing-core",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


dvdplm:
> * Aff-g* is _super_ slow. Takes 13 seconds to generate a single proof. Is it supposed to be like that? We only do it in the error round in Presigning, but still.

Yeah, that's what I see too. Verification is also very slow:

```
Benchmarking AffG* proof/prove: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 122.0s.
AffG* proof/prove       time:   [11.964 s 12.042 s 12.141 s]
                        change: [-2.8507% -1.8706% -0.7880%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
Benchmarking AffG* proof/verify: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 232.5s.
AffG* proof/verify      time:   [10.791 s 10.854 s 10.945 s]
                        change: [+0.3281% +0.9628% +1.8262%] (p = 0.01 < 0.05)
                        Change within noise threshold.
```

I took a few samples while the benchmark is running, both during proving and verification and there's nothing funny going on in the benchmark themselves afaict: it *is* spending the time in side `AffGStarProof::new` and `AffGStarProof::verify`.

dvdplm (commented):
The benchmarks are now more custom than they were before, that's a downside. The point of the previous setup was that it was pretty easy for anyone used to reading Criterion benchmarks to read. With this PR it there's more code to understand.

The motivation for this refactor was – I believe – to speed up the overall runtime by avoiding duplicate setup code. What was the actual outcome, how long does a benchmark run take now vs before?

fjarri:
I don't think there's more code - there's explicitly 500 lines less. And anyone reading the benchmarks would still have to go to `private_benches.rs` to look at the actual functions - this part has not changed. 

The goals were:
- to move `criterion` out of the crate and into the benchmark executable
- to simplify the code in `private_benches`

The draft of this PR contained a variant where I relied on `criterion` a little more, exposing stuff to run in `iter_batched` instead of `iter_custom`, but that required returning closures and repeating secret/public parameters (see [this commit](https://github.com/entropyxyz/synedrion/commit/c8cec19610fe4ae3a82ffd69273fea60ddfaf154#diff-2ce6434549d7bc7cb4534caf48e3ec4f62bfb6b680eaa170ae4d3b7726f7c770R52)). I was delaying this PR for a month trying to decide if that approach is worth it, but in the end went with the simpler thing. 

If we keep `criterion` in the main dependencies (`private-benches` is already excluded from the feature powerset test, so maybe it'll be fine), why not just expose `bench_<..>(&mut Criterion)` from `private_benches.rs`? Then there would be no awkward separation of logic between two files, and no closures need to be returned.

dvdplm:
> why not just expose `bench_<..>(&mut Criterion)` from `private_benches.rs`? Then there would be no awkward separation of logic between two files, and no closures need to be returned.

Yeah maybe that's worth exploring actually.

I think having access to criterion is very useful.

Issue #187, opened 2025-02-04
Handle out-of-range `\alpha`/`\hat{\alpha}` in Presigning


fjarri:
Extracted from the discussion in #181

In Presigning (Fig. 8), Round 3, we decrypt `\alpha_{i,j}` and use them to construct `\delta_i` (the following issue applies similarly to `\hat{\alpha}_{i,j}` and `\chi_i` as well). In the happy path `\delta_i` is reduced `mod q`, so there are no range issues.

But if there is a problem, and we have to go into the error round (Fig. 9), we need to create a `П^{dec}` based on the non-reduced value of `\delta_i`. The proof requires it to be in `±2^{\ell^\prime}`; if it's not, the later proof verification will fail. So if some node sends an `\alpha_{i,j}` that's out of the expected range, it will cause us to incriminate ourselves by sending out an invalid proof.

As a matter of fact, it's not `±2^{\ell^\prime}` even in the happy path — it's a sum of `2(n-1)` random `beta` variables, where `n` is the number of parties, so it's technically something like `±2^{\ell^\prime + log(n) + 1}`.

How can we deal with that? Possible solutions include:
- On receiving `D_{i,j}` and decrypting it into `\alpha_{i,j}`, check it is in the correct range. If not, publish the evidence with the message containing `D_{i,j}`, adding the revealed `\alpha_{i,j}` and the randomizer. Questions: 1) is that a sufficient proof (that is, are we sure it can't be forged)?; 2) is it safe to reveal `\alpha_{i,j}` assuming the session will be halted? I *think* it's yes to both, but I'm not sure.
- Send out some kind of a range proof along with `D_{i,j}`

PR #193, opened 2025-03-05
TestParams use TinyCurve32

dvdplm:
Adjust test params to be as tight as possible:

- use TinyCurve32
- `L_BOUND` is 32
- `EPS_BOUND` is 64 (2*32)
- `LP_BOUND` is 160 (5*32)
- `PRIME_BITS` is 128 (so Uint is U256 etc)

The speedup when running tests is dramatic.

Master:

```
hyperfine -w 2 "cargo t --release"
 Benchmark 1: cargo t --release
   Time (mean ± σ):      7.790 s ±  0.122 s    [User: 43.535 s, System: 0.446 s]
   Range (min … max):    7.632 s …  8.011 s    10 runs
```

This PR:

```
hyperfine -w 2 "cargo t --release"
Benchmark 1: cargo t --release
  Time (mean ± σ):     942.8 ms ±  18.9 ms    [User: 2882.5 ms, System: 166.0 ms]
  Range (min … max):   919.8 ms … 971.8 ms    10 runs
```

**NOTE**: There is one potentially problematic change in this PR, to be reviewed carefully. The `checked_add` impl for `SecretSigned<T>` used to pick the biggest bound from the two operands and *then add 1* to it. This leads to a few test failures with the params as configured in this PR. The underlying reason is that now the curve order is so much closer to uint sizes that we run into corner cases where e.g. [Paillier decryption yields a SecretSigned with a bound of 255](https://github.com/entropyxyz/synedrion/blob/dp-use-TinyCurve32-in-tests/synedrion/src/paillier/encryption.rs#L224) (which comes from `MODULUS_BITS`), causing addition to always "overflow" (actually: fail the bounds check that `bound + 1 < T::BITS`).

~~We have discussed this bound increment last summer and I think we decided to keep the increment as the conservative choice, but I actually think it's wrong. Maybe now is the right time to make the final call on this?~~

I'm wrong, the last bit must be preserved. However, the simple bounds check breaks down in this corner case and we need to do something smarter for this to work. How about checking the last bit before and after the sum, and if the bit flipped we overflowed?

### DIFF
--- a/synedrion/src/cggmp21/params.rs
+++ b/synedrion/src/cggmp21/params.rs
@@ -5,6 +5,7 @@ use core::{fmt::Debug, ops::Add};
 // So as long as that is the case, `k256` `Uint` is separate
 // from the one used throughout the crate.
 use crypto_bigint::{BitOps, NonZero, Uint, U1024, U2048, U4096, U512, U8192};
+use crypto_bigint::{U128, U256};
 use digest::generic_array::{ArrayLength, GenericArray};
 use ecdsa::hazmat::{DigestPrimitive, SignPrimitive, VerifyPrimitive};
 use primeorder::elliptic_curve::{
@@ -15,12 +16,12 @@ use primeorder::elliptic_curve::{
 };
 use serde::{Deserialize, Serialize};
 
-use tiny_curve::TinyCurve64;
+use tiny_curve::TinyCurve32;
 
 use crate::{
     paillier::PaillierParams,
     tools::hashing::HashableType,
-    uint::{U1024Mod, U2048Mod, U4096Mod, U512Mod},
+    uint::{U1024Mod, U128Mod, U2048Mod, U256Mod, U4096Mod, U512Mod},
 };
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -85,14 +86,14 @@ impl PaillierParams for PaillierTest {
     are much smaller than 2*PRIME_BITS.
     */
 
-    const PRIME_BITS: u32 = 397;
-    type HalfUint = U512;
-    type HalfUintMod = U512Mod;
-    type Uint = U1024;
-    type UintMod = U1024Mod;
-    type WideUint = U2048;
-    type WideUintMod = U2048Mod;
-    type ExtraWideUint = U4096;
+    const PRIME_BITS: u32 = 128;
+    type HalfUint = U128;
+    type HalfUintMod = U128Mod;
+    type Uint = U256;
+    type UintMod = U256Mod;
+    type WideUint = U512;
+    type WideUintMod = U512Mod;
+    type ExtraWideUint = U1024;
 }
 
 /// Paillier parameters corresponding to 112 bits of security.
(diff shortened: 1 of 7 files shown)


fjarri:
That's strange, where exactly in the protocol does the overflow occur? There shouldn't be any problems as long as the modulus size is larger than, roughly `q^2 * n` where `q` is the curve order and `n` is the number of nodes, and we're still way beyond that even with test parameters. It might be that we're actually catching a bug in the logic sopmewhere.

dvdplm:
> where exactly in the protocol does the overflow occur?

So first of all, let me be clear about what "overflow" means here: it's actually not that the case that there's arithmetic that *actually* overflows with these new params. What happens is that the bounds check in `impl Add<SecretSigned<T>> for SecretSigned<T>` checks that the biggest bound of the operands **plus one** is smaller than `T::BITS`. 
Before this PR, `T::BITS` was 2048 and the numbers involved were bounded on the curve modulus or `LP_BOUND`/`EPS_BOUND` so much much smaller. With this PR, the uint sizes used and the curve order (which determines `MODULUS_BITS`, `LP_BOUND` and `EPS_BOUND`) are much closer so the "coarse" bounds check now matters much more.

Take [`r5_dec_failed`](https://github.com/entropyxyz/synedrion/blob/dp-use-TinyCurve32-in-tests/synedrion/src/cggmp21/interactive_signing_tests.rs#L511) for instance. There's an "overflow" when adding `beta` and `secret.y` [here](https://github.com/entropyxyz/synedrion/blob/dp-use-TinyCurve32-in-tests/synedrion/src/cggmp21/sigma/dec.rs#L151), because `secret.y` has a [bound of 255](https://github.com/entropyxyz/synedrion/blob/dp-use-TinyCurve32-in-tests/synedrion/src/paillier/encryption.rs#L224) (`beta` has a bound of `LP_BOUND + EPS_BOUND = 224` but that doesn't matter here). The "overflow" happens [because `255 + 1` is not less than `T::BITS` (256)](https://github.com/entropyxyz/synedrion/blob/master/synedrion/src/uint/secret_signed.rs#L207).

When I say that the current "overflow" check is too coarse, what I mean is a `SecretSigned<U256>` with bound 255 will always overflow, regardless of what numbers are involved. Is that ok? Why not actually check if there's an overflow, so that `SecretSigned::from(value: U256::ZERO, bound: 255) + SecretSigned::from(value: U256::ONE, bound: 123)` can work (pseudocode)?

fjarri:
But that's specifically #187. It's not caused by the bound propagation, there still should be plenty of space. It's a drawback of the scheme which should be resolved on the "theoretical" level, then the implementation level will be correct. 

As for what to do with it right now, I'm not sure - need to think about it.

dvdplm:
> As for what to do with it right now, I'm not sure

Seems like we shouldn't rush this. I'll back out my changes to `checked_add` and loosen the params to make it work. That way we can merge this and enjoy the fast tests while we ponder what we do here.

dvdplm:
Opened #195 that I think we can merge pretty easily.

fjarri:
Overridden by #195

PR #195, opened 2025-03-07
Use TinyCurve32 for tests

dvdplm:
A replacement PR for https://github.com/entropyxyz/synedrion/pull/193 that requires some careful thought.

This PR uses `TinyCurve32` just like #193 but sets `PRIME_BITS` to the (wrong) value of 127, which avoids the test failures but is still about as correct as the previous TestParams.


fjarri:
I will merge this for now and fix #187 in the follow-up.

PR #197, opened 2025-03-08
Ensure correct size of `y` argument for `П^{dec}`

### LINKED ISSUE #187: Handle out-of-range `\alpha`/`\hat{\alpha}` in Presigning

### LINKED ISSUE #189: Add scheme parameters corresponding to 128 bits of security
See Section C.1 in the paper. Currently we have 112-bit ones implemented. The 128 bit parameters require a 12288-bit `Uint`, which doesn't have a predefined alias and all the corresponding impls (like `Concat`). We can implement those manually, or perhaps use `BoxedUint`?.

fjarri:
- Change the modulus size in `TestParams` to 128. 
- Ensure the correct size of `y` argument for `П^{dec}` (`delta_i` and `chi_i`). See the change in `PAPER.md` for the theoretical background. Most of the changes are just dealing with the `П^{dec}` now requiring some operations on wide Uints.

I think this is the correct approach from the mathematical perspective (although having to add some `_wide` methods is not great, see also #198). Fixes #187. I think it is safe to do since we added the `DEVIATION` comments in the proof itself, and the node in `PAPER.md`, so the future audit will review it.

### DIFF
--- a/PAPER.md
+++ b/PAPER.md
@@ -46,6 +46,22 @@ The above item leads to another problem. If those values are indeed echo-broadca
 Fig. 8, Round 2, 2b) - `\psi_{j,i}` creation requires `F_{i,j}` which are not yet available since they're the ones created on other nodes. The previous paper version has `F_{j,i}` there. Same for `\hat{\psi_{j,i}}`.
 
 
+# Range of `delta_i`
+
+There is some inconsistency with ranges in the Presigning protocol (Fig. 8).
+
+When we construct `delta_i` in Round 3, step 2, its range is `±2^{\ell^\prime + \eps + 1 + ceil(log2(n))}` where `n` is the number of parties.
+`\ell^\prime + \eps` is the range of a single `beta_{j,i}` coming from another node (as proven by `П^{aff-g}` in step 1); all the other terms have much smaller ranges, so adding them results in increasing the range by at most 1 bit.
+Since there are `n` `beta_{j,i}` in the sum, we add `ceil(log2(n))`, getting the final expression.
+
+If we reach the error round (Fig. 9) and need to construct `П^{dec}`, `delta_i` serves as the `y` argument.
+But the description of the proof (Fig. 28) requires `y ∈ ±2^{\ell^\prime}` (and correspondingly, `\beta` is sampled from `±2^{\ell^\prime + \eps}` to mask `y`).
+
+So to accommodate for the actual range of `delta_i`, we have to bump the range of `beta` and the range check of `w` by `\eps + 1 + ceil(log2(n))`, and add the number of parties to the public parameters of `П^{dec}`.
+
+Same applies to `chi_i`.
+
+
 # Echo-broadcasting
 
 In order to be able to generate verifiable evidence for each failure, some values have to be echo-broadcasted instead of normal broadcast/direct messaging given in the paper. Here is the list of such variables
--- a/synedrion/src/cggmp21/conversion.rs
+++ b/synedrion/src/cggmp21/conversion.rs
@@ -6,7 +6,7 @@ use crate::{
     curve::Scalar,
     paillier::PaillierParams,
     tools::Secret,
-    uint::{PublicSigned, SecretSigned},
+    uint::{HasWide, PublicSigned, SecretSigned},
 };
 
 fn uint_from_scalar<P: SchemeParams>(value: &Scalar<P>) -> <P::Paillier as PaillierParams>::Uint {
@@ -37,8 +37,8 @@ pub(crate) fn public_signed_from_scalar<P: SchemeParams>(
 }
 
 /// Converts an integer to the associated curve scalar type.
-pub(crate) fn scalar_from_uint<P: SchemeParams>(value: &<P::Paillier as PaillierParams>::Uint) -> Scalar<P> {
-    let r = *value % P::CURVE_ORDER;
+pub(crate) fn scalar_from_wide_uint<P: SchemeParams>(value: &<P::Paillier as PaillierParams>::WideUint) -> Scalar<P> {
+    let r = *value % P::CURVE_ORDER_WIDE;
 
     let repr = r.to_be_bytes();
     let uint_len = repr.as_ref().len();
@@ -57,7 +57,18 @@ pub(crate) fn scalar_from_uint<P: SchemeParams>(value: &<P::Paillier as Paillier
 pub(crate) fn scalar_from_signed<P: SchemeParams>(
     value: &PublicSigned<<P::Paillier as PaillierParams>::Uint>,
 ) -> Scalar<P> {
-    let abs_value = scalar_from_uint::<P>(&value.abs());
+    let abs_value = scalar_from_wide_uint::<P>(&value.abs().to_wide());
+    if value.is_negative() {
+        -abs_value
+    } else {
+        abs_value
+    }
+}
+
+pub(crate) fn scalar_from_wide_signed<P: SchemeParams>(
+    value: &PublicSigned<<P::Paillier as PaillierParams>::WideUint>,
+) -> Scalar<P> {
+    let abs_value = scalar_from_wide_uint::<P>(&value.abs());
     if value.is_negative() {
         -abs_value
     } else {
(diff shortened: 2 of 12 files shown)


synedrion/src/paillier/encryption.rs
```diff
@@ -119,8 +108,9 @@ impl<P: PaillierParams> Ciphertext<P> {
         // Since `m` can be negative, we calculate `m * N ± 1` (never overflows since `m < N`),
         // then conditionally negate modulo N^2
 
-        let prod = abs_plaintext.mul_wide(pk.modulus());
-        let mut prod_mod = prod.to_montgomery(pk.monty_params_mod_n_squared());
+        let abs_plaintext = abs_plaintext.to_montgomery(pk.monty_params_mod_n_squared());
```
fjarri, line 111: This decreases performance somewhat, but makes it possible to use this method both for `Uint` and `WideUint`. Worth it?

synedrion/src/cggmp21/conversion.rs
```diff
@@ -66,10 +77,10 @@ pub(crate) fn scalar_from_signed<P: SchemeParams>(
 }
 
 /// Converts a secret-wrapped uint to a secret-wrapped [`Scalar`], reducing the value modulo curve order.
-fn secret_scalar_from_uint<P: SchemeParams>(
-    value: &Secret<<P::Paillier as PaillierParams>::Uint>,
+fn secret_scalar_from_wide_uint<P: SchemeParams>(
+    value: &Secret<<P::Paillier as PaillierParams>::WideUint>,
 ) -> Secret<Scalar<P>> {
-    let r = value % &P::CURVE_ORDER;
+    let r = value % &P::CURVE_ORDER_WIDE;
```
dvdplm, line 83: Maybe it's possible to use `RemMixed` here and get a smaller `r` right away. Probably not worth the hassle though?

PAPER.md
```diff
@@ -46,6 +46,22 @@ The above item leads to another problem. If those values are indeed echo-broadca
 Fig. 8, Round 2, 2b) - `\psi_{j,i}` creation requires `F_{i,j}` which are not yet available since they're the ones created on other nodes. The previous paper version has `F_{j,i}` there. Same for `\hat{\psi_{j,i}}`.
 
 
+# Range of `delta_i`
+
+There is some inconsistency with ranges in the Presigning protocol (Fig. 8).
+
+When we construct `delta_i` in Round 3, step 2, its range is `±2^{\ell^\prime + \eps + 1 + ceil(log2(n))}` where `n` is the number of parties.
```
dvdplm, line 53: Nitpick: `\eps` should be `\epsilon` to be valid LaTex I think?

dvdplm:
FWIW this PR does not seem to impact benchmark performance in any significant way.

dvdplm (approved):

synedrion/src/cggmp21/params.rs
```diff
@@ -174,14 +130,6 @@ where
...
 pub struct TestParams;
 
-// Some requirements from range proofs etc:
-// - $П_{enc}$, safe two's complement representation of $\alpha$ requires
-//   `L_BOUND + EPS_BOUND + 1 < Uint::BITS - 1`
-// - $П_{enc}$, safe two's complement representation of $z_1$ requires
-//   `L_BOUND + max(EPS_BOUND, log2(q)) + 1 < Uint::BITS - 1`
-//   (where `q` is the curve order)
-// - Range checks will fail with the probability $q / 2^\eps$, so $\eps$ should be large enough.
-// - P^{fac} assumes $N ~ 2^{4 \ell + 2 \eps}$
```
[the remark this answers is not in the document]
fjarri: It was useful when we had to juggle the bounds to allow all the proofs to work, but if we choose the ratios recommended by the paper there's plenty of space left. What is it specifically useful for? Perhaps it can be disbanded and moved to appropriate places?

PAPER.md
```diff
@@ -46,6 +46,22 @@ The above item leads to another problem. If those values are indeed echo-broadca
...
+`\ell^\prime + \eps` is the range of a single `beta_{j,i}` coming from another node (as proven by `П^{aff-g}` in step 1); all the other terms have much smaller ranges, so adding them results in increasing the range by at most 1 bit.
+Since there are `n` `beta_{j,i}` in the sum, we add `ceil(log2(n))`, getting the final expression.
+
+If we reach the error round (Fig. 9) and need to construct `П^{dec}`, `delta_i` serves as the `y` argument.
+But the description of the proof (Fig. 28) requires `y ∈ ±2^{\ell^\prime}` (and correspondingly, `\beta` is sampled from `±2^{\ell^\prime + \eps}` to mask `y`).
+
+So to accommodate for the actual range of `delta_i`, we have to bump the range of `beta` and the range check of `w` by `\eps + 1 + ceil(log2(n))`, and add the number of parties to the public parameters of `П^{dec}`.
+
```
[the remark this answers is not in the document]
fjarri: What public params specifically? You mean, to the bounds in `SchemeParameters`? I'm not sure that would work, since this addition is only needed here.

synedrion/src/cggmp21/conversion.rs
fjarri: It's only implemented up to `U1024`. I think it would be much more useful to have access to `rem_vartime()` here, but it doesn't have an associated trait.

PAPER.md
fjarri: Oh yes, you're right (I think it's `\varepsilon` actually). Haven't written in TeX for so long I start forgetting the aliases :)

synedrion/src/cggmp21/conversion.rs
fjarri: Filed https://github.com/entropyxyz/synedrion/issues/201

PAPER.md
fjarri: Filed https://github.com/entropyxyz/synedrion/issues/202

PAPER.md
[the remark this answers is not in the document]
fjarri: Oh wait, you mean the public parameters of `dec`. Yes, we could, but the number of parties is what's available to the caller, and if `dec` needs to take a logarithm of it, it's its internal business.

synedrion/src/cggmp21/conversion.rs
dvdplm: > I think it would be much more useful to have access to `rem_vartime()` here, but it doesn't have an associated trait.

`RemMixed` is impl'd with `div_rem_vartime`, which is what `rem_vartime` uses too, so I think we're saying the same thing?

fjarri: > `RemMixed` is impl'd with `div_rem_vartime`

That's actually quite unexpected. I don't think it should use vartime.

dvdplm: I named the trait `RemVartime` originally, but that's not a great name either given it operates on mixed input. I guess `RemMixedVartime` would have conveyed the point fully but it's a bit of a mouthful.

PR #199, opened 2025-03-09
Cleanups and feature gating

fjarri:
- remove `primeorder` (all we need is in `elliptic-curve`), `bincode`, and `sha2` dependencies
- create a `k256` feature and put all the `k256` stuff behind it (including production parameters)
- create a `dev` feature and put test parameters behind it
- create a `bip32` feature and put all the `bip32` stuff behind it
- had to move the integration test to the `src/tests` module because of Cargo's inability to specify required features for integration tests
- cleaned up unused features from dependencies
- added a `cargo hack` CI action to build with different feature sets

The feature powerset test is the last to finish, which is kind of annoying. We'll see, maybe we'll remove it later.


.github/workflows/ci.yml
```diff
@@ -27,8 +27,25 @@ jobs:
...
-      - run: cargo build --target ${{ matrix.target }}
+      - uses: taiki-e/install-action@cargo-hack
+      - run: cargo hack build --target ${{ matrix.target }} --feature-powerset
+
+  build-feature-powerset:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.83.0
```
dvdplm, line 38: We'll want 1.85 so we get the 2024 edition perhaps?

dvdplm (approved):
I wanted to do something similar but got a bit stuck on whether we want a default feature or not (that's be `k256` + `bip32`), but I guess that's easy enough to set in `entropy-tss`.

fjarri:
I'm not a fan of default features in general, I'm starting to think they were a design mistake in `cargo`.

.github/workflows/ci.yml
fjarri: `crypto-bigint=0.6` has MSRV 1.83, we'll bump when we switch to 0.7.

PR #200, opened 2025-03-10
"ungroup" slow tests into individual tests for faster execution

dvdplm:
When using `nextest` to run tests, this PR cuts total running time from 8 to 4 sec, by splitting up the slowest tests into smaller units such that nextest's "test-per-process" model can execute them concurrently.


fjarri:
Looks good to me, perhaps it can be extended to the other protocols? And maybe we should separate the tests for each round too?

PR #203, opened 2025-03-11
Flatten module hierarchy

fjarri:
- Get rid of the multirepo. We are not planning to add any other crates here.
- Move ZK proofs to the `zk` top-level module (`sigma` is a term specific to CGGMP, "ZK" is more generally understood)
- Put all the protocols in the `protocols` module (the crate declares that it's an implementation to CGGMP, so there's no need to call module that explicitly; and WWW'02 is really just an extension of CGGMP)
- Put all entities (key shares and so on) into the top-level `entities` module
- Put all scheme params and related things into the `params` top-level module
- Move malicious behavior tests to a submodule


fjarri:
The difference in coverage is because prior to the commit moving the protocol tests, their code was included in the coverage. Seems to be some quirk of `llvm-cov` because the code in the inline `mod tests` is still included, which seems inconsistent.

PR #204, opened 2025-03-12
Generic hash

fjarri:
- Make the hasher an associated type of `SchemeParams::Digest`, since we want the hash security to match overall security.
- Renamed `XofHasher` to `Hasher`.
- Wrapped hash output, SID, and EPID into their own types
- Hashing all scheme parameters, not just the curve order

### DIFF
--- a/src/entities/internal.rs
+++ b/src/entities/internal.rs
@@ -18,7 +18,7 @@ impl Sid {
                 .chain_type::<P::Curve>()
                 .chain(&shared_randomness)
                 .chain(&ids)
-                .finalize_boxed(P::SECURITY_BITS),
+                .finalize(),
         )
     }
 }
--- a/src/entities/internal.rs
+++ b/src/entities/internal.rs
@@ -4,7 +4,7 @@ use manul::protocol::PartyId;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    params::SchemeParams,
+    params::{chain_scheme_params, SchemeParams},
     tools::hashing::{Chain, HashOutput, Hasher},
 };
 
@@ -16,12 +16,10 @@ pub(crate) struct Sid(HashOutput);
 
 impl Sid {
     pub fn new<P: SchemeParams, Id: PartyId>(shared_randomness: &[u8], ids: &BTreeSet<Id>) -> Self {
-        Self(
-            Hasher::<P::Digest>::new_with_dst(b"SID")
-                .chain_type::<P::Curve>()
-                .chain(&shared_randomness)
-                .chain(&ids)
-                .finalize(P::SECURITY_BITS),
-        )
+        let digest = Hasher::<P::Digest>::new_with_dst(b"SID");
+        let digest = chain_scheme_params::<P, _>(digest);
+        let digest = digest.chain(&shared_randomness).chain(&ids);
+
+        Self(digest.finalize(P::SECURITY_BITS))
     }
 }
--- a/src/entities/internal.rs
+++ b/src/entities/internal.rs
@@ -8,6 +8,9 @@ use crate::{
     tools::hashing::{Chain, HashOutput, Hasher},
 };
 
+/// The session identifier (see Remark 4.1 in the paper).
+///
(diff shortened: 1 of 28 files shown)


src/tools/hashing.rs
```diff
@@ -46,27 +48,30 @@ impl Chain for XofHasher {
...
+    pub fn finalize(self) -> HashOutput {
         // A common heuristic for hashes is that the log2 of the collision probability is half the output size.
         // We may not have enough output bytes, but this constitutes the best effort.
-        self.0.finalize_xof().read_boxed((security_bits * 2).div_ceil(8))
+        HashOutput(self.0.finalize_xof().read_boxed((P::SECURITY_BITS * 2).div_ceil(8)))
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HashOutput(#[serde(with = "SliceLike::<Hex>")] Box<[u8]>);
```
dvdplm, line 73: I think that the previous code being decoupled from `SchemeParams` was a good thing and that the return of `HashOutput` is a "downgrade".

src/protocols/key_refresh.rs
```diff
@@ -517,12 +504,7 @@ impl<P: SchemeParams, Id: PartyId> EntryPoint<Id> for KeyRefresh<P, Id> {
 
         let other_ids = self.all_ids.clone().without(id);
 
-        let sid = make_sid::<P, Id>(
-            shared_randomness,
-            &KeyRefreshAssociatedData {
-                ids: self.all_ids.clone(),
-            },
-        );
+        let sid = Sid::new::<P, Id>(shared_randomness, &self.all_ids);
```
dvdplm, line 507: :+1

src/entities/internal.rs
```diff
@@ -0,0 +1,24 @@
...
+use manul::protocol::PartyId;
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    params::SchemeParams,
+    tools::hashing::{Chain, HashOutput, Hasher},
+};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct Sid(HashOutput);
```
dvdplm, line 12:
```suggestion
/// The session identifier (see Remark 4.1 in the paper).
#[derive(Debug, Clone, Serialize, Deserialize)]
pub(crate) struct Sid(HashOutput);
```

dvdplm (approved):
I think this is better, but I'm still a bit irked by having to lug around more generics and still have to provide the desired length (semi-)manually.

src/tools/hashing.rs
fjarri: The wrapper is there mainly to encapsulate the serialization attribute. 

As for the hasher being a part of scheme parameters, while it adds to the complexity somewhat, I think it is a logical  and, I guess, inevitable consequence of `SECURITY_BITS` being a user-defined parameter.

src/params/traits.rs
```diff
@@ -56,6 +57,11 @@ where
     /// Double the curve Scalar-width integer type.
     type WideCurveUint: bigintv05::Integer + Split<Output = <Self::Curve as Curve>::Uint>;
 
+    /// The hash that will be used for protocol's internal purposes.
+    ///
+    /// Note: the collision probability must be consistent with [`Self::SECURITY_BITS`].
+    type Digest: Default + Update + ExtendableOutput;
```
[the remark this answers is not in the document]
fjarri: We're not really using `FixedOutput`, and `Shake256` doesn't even implement it. Would be nice to have collision probability as an associated constant, but I don't think there's anything like that for XOF hashes.

src/tools/hashing.rs
fjarri: Your intuition was right, being generic over parameters didn't really work out, which I realized when I added hashing all the scheme parameters. Hashing Paillier parameters using the scheme as the generic would be weird.

dvdplm:
@fjarri I like the new commits. My approval still stands.

PR #208, opened 2025-03-20
Check consistency between 32- and 64-bit platforms

fjarri:


src/uint/traits.rs
```diff
@@ -220,3 +220,31 @@ impl<const L: usize> BoxedEncoding for Uint<L> {
...
+
+#[cfg(test)]
+mod tests {
+    use crypto_bigint::U256;
+    use sha3::Shake256;
+
+    use super::FromXofReader;
+    use crate::tools::hashing::{Chain, Hasher};
+
+    #[test]
```
dvdplm, line 232:
```suggestion
    // Test to ensure that random seeds work the same on 32 and 64 bit platforms
    #[test]
```

…or something to that effect.

dvdplm (commented):
I think the tests in rsa.rs and traits.rs both need a comment explaining where the numbers come from and why the tests are there in the first place.

LGTM otherwise (although: should we bump MSRV?)

PR #213, opened 2025-04-03
Handle secrets in constant time

dvdplm:
Make most (all?) methods in `SecretSigned` and `SecretUnsigned` be constant time where possible.

### DIFF
--- a/src/uint/secret_signed.rs
+++ b/src/uint/secret_signed.rs
@@ -1,4 +1,4 @@
-use core::ops::{Add, Mul, Neg, Sub};
+use core::ops::{Add, BitAnd, Mul, Neg, Not, Sub};
 
 use crypto_bigint::{
     rand_core::CryptoRngCore,
@@ -33,18 +33,11 @@ where
 
     /// Creates a signed value from an unsigned one, assuming that it encodes a positive value
     /// treated as two's complement.
-    ///
-    /// Panics if it is not the case.
-    pub fn new_positive(value: Secret<T>, bound: u32) -> Option<Self> {
-        // Reserving one bit as the sign bit (MSB)
-        if bound >= T::BITS || value.expose_secret().bits() > bound {
-            return None;
-        }
-        let result = Self::new_from_unsigned_unchecked(value, bound);
-        if result.is_negative().into() {
-            return None;
-        }
-        Some(result)
+    pub fn new_positive(value: Secret<T>, bound: u32) -> CtOption<Self> {
+        let in_bound = Choice::from((bound < T::BITS && value.expose_secret().bits() <= bound) as u8);
+        let new_positive = Self::new_from_unsigned_unchecked(value, bound);
+        let is_positive = new_positive.is_negative().not();
+        CtOption::new(new_positive, in_bound.bitand(is_positive))
     }
 
     pub fn zero() -> Self {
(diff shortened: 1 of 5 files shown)


src/uint/secret_signed.rs
```diff
@@ -33,18 +33,11 @@ where
...
-        // Reserving one bit as the sign bit (MSB)
-        if bound >= T::BITS || value.expose_secret().bits() > bound {
-            return None;
-        }
-        let result = Self::new_from_unsigned_unchecked(value, bound);
-        if result.is_negative().into() {
-            return None;
-        }
-        Some(result)
+    pub fn new_positive(value: Secret<T>, bound: u32) -> CtOption<Self> {
```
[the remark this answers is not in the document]
fjarri: Sort of clashes with other places where we return `Choice` though. I think we need to pick either `subtle` or the const-time `crypto-bigint` stuff and use that.

src/uint/secret_signed.rs
```diff
@@ -57,7 +50,7 @@ where
     /// Returns a truthy `Choice` if this number is negative.
     pub fn is_negative(&self) -> Choice {
         // Check the MSB, `1` indicates that it is negative.
-        Choice::from(self.value.expose_secret().bit_vartime(T::BITS - 1) as u8)
+        self.value.expose_secret().bit(T::BITS - 1)
```
fjarri, line 53: we can keep `bit_vartime()` here, since the bit place is constant

src/uint/secret_signed.rs
```diff
@@ -97,21 +90,19 @@ where
     /// Creates a [`SignedSecret`] from an unsigned value, treating it as if it encodes the sign as two's complement.
     ///
     /// Returns `None` if the requested bound is too large, or if `abs(value)` is actually larger than the bound.
-    pub fn new_from_unsigned(value: Secret<T>, bound: u32) -> Option<Self> {
-        let is_negative = Choice::from(value.expose_secret().bit_vartime(T::BITS - 1) as u8);
+    pub fn new_from_unsigned(value: Secret<T>, bound: u32) -> CtOption<Self> {
+        let is_negative = value.expose_secret().bit(T::BITS - 1);
```
fjarri, line 94: `bit_vartime()` can be kept there as well

src/uint/secret_signed.rs
```diff
@@ -97,21 +90,19 @@ where
...
         let abs = Secret::<T>::conditional_select(&value, &value.wrapping_neg(), is_negative);
         // Reserving one bit as the sign bit (MSB)
-        if bound >= T::BITS || abs.expose_secret().bits() > bound {
-            return None;
-        }
-        Some(Self::new_from_unsigned_unchecked(value, bound))
+        let in_bound = Choice::from((bound < T::BITS && abs.expose_secret().bits() <= bound) as u8);
```
fjarri, line 97: Need to document that it's variable-time w.r.t. `bound`

src/uint/secret_signed.rs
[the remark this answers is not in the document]
dvdplm: I think you might be confusing `subtle::Choice` and `crypto_bigint::ConstChoice` here yes? The `subtle::CtOption` replaces Rust's `Option`.

I don't think we're using `ConstChoice` anywhere in `synedrion`?

src/uint/secret_signed.rs
dvdplm: Not sure what you mean here: after this PR it is not variable time in the bound, so not sure what to document!

fjarri:
I am not sure if this is actually necessary. The methods returning `Option` were only used in the assertions, so this doesn't change how much information is leaked. Is it a just in case change, if we use them somewhere else in the future?

src/uint/secret_signed.rs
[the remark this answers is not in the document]
fjarri: Wouldn't the comparisons and logic ops have to be made constant-time too?

PR #214, opened 2025-04-04
Save some clones while interpolating polynomial coefficients

dvdplm:
I think it's more efficient to leverage BTreeMaps unique keys and avoid building BTreeSets when evaluating interpolation coefficients.

### DIFF
--- a/src/entities/threshold.rs
+++ b/src/entities/threshold.rs
@@ -161,10 +161,9 @@ where
                     .ok_or_else(|| LocalError::new("id={id:?} is missing in the share_ids"))?;
                 Ok((id.clone(), *share_id))
             })
-            .collect::<Result<BTreeMap<_, _>, LocalError>>()?;
+            .collect::<Result<BTreeMap<I, ShareId<P>>, LocalError>>()?;
 
-        let share_ids_set = share_ids.values().cloned().collect();
-        let secret_share = self.secret_share.clone() * interpolation_coeff(&share_ids_set, owner_share_id);
+        let secret_share = self.secret_share.clone() * interpolation_coeff(share_ids.values(), owner_share_id);
         let public_shares = ids
             .iter()
             .map(|id| {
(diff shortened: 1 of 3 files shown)


src/tools/sss.rs
```diff
@@ -84,7 +68,13 @@ where
     }
 
     pub fn evaluate(&self, x: &ShareId<P>) -> Secret<Scalar<P>> {
-        evaluate_polynomial_secret(&self.0, &x.0)
+        // Evaluate in reverse to save on multiplications.
```
dvdplm, line 71: Inlined the code here to make things easier to follow.

src/tools/sss.rs
```diff
@@ -162,10 +153,9 @@ pub(crate) fn shamir_join_points<P>(pairs: &BTreeMap<ShareId<P>, Point<P>>) -> P
 where
     P: SchemeParams,
 {
-    let share_ids = pairs.keys().cloned().collect::<BTreeSet<_>>();
     pairs
         .iter()
-        .map(|(share_id, val)| val * interpolation_coeff(&share_ids, share_id))
+        .map(|(share_id, val)| val * interpolation_coeff(pairs.keys(), share_id))
```
dvdplm, line 158: I think this approach is more lightweight yeah?

src/entities/threshold.rs
```diff
@@ -161,10 +161,9 @@ where
                     .ok_or_else(|| LocalError::new("id={id:?} is missing in the share_ids"))?;
                 Ok((id.clone(), *share_id))
             })
-            .collect::<Result<BTreeMap<_, _>, LocalError>>()?;
+            .collect::<Result<BTreeMap<I, ShareId<P>>, LocalError>>()?;
```
dvdplm, line 164: This is just to help rust-analyzer resolve the types in the IDE (and perhaps it's helping the casual reader too?).

PR #215, opened 2025-04-07
Bump `manul` to 0.2.0

fjarri:


src/curve/arithmetic.rs
```diff
@@ -66,12 +66,12 @@ impl<P: SchemeParams> Scalar<P> {
...
-    pub fn random(rng: &mut impl CryptoRngCore) -> Self {
-        Self(ScalarPrimitive::<P::Curve>::random(rng).into())
+    pub fn random(rng: &mut dyn CryptoRngCore) -> Self {
+        Self(ScalarPrimitive::<P::Curve>::random(&mut BoxedRng(rng)).into())
     }
 
-    pub fn random_nonzero(rng: &mut impl CryptoRngCore) -> Self {
-        Self(*NonZeroScalar::<P::Curve>::random(rng).as_ref())
+    pub fn random_nonzero(rng: &mut dyn CryptoRngCore) -> Self {
+        Self(*NonZeroScalar::<P::Curve>::random(&mut BoxedRng(rng)).as_ref())
```
dvdplm, line 74: This is pretty annoying esthetically; it's pretty much just doing `Uint::random_mod(&mut rng, P::CURVE_ORDER)`. Sigh.

src/tools/boxed_rng.rs
```diff
@@ -0,0 +1,23 @@
+use rand_core::{CryptoRng, CryptoRngCore, RngCore};
+
+/// Since object-safe trait methods cannot take `impl CryptoRngCore` arguments,
+/// this structure wraps the dynamic object and exposes a `CryptoRngCore` interface,
+/// to be passed to statically typed round methods.
```
dvdplm, line 5:
```suggestion
/// Since object-safe trait methods cannot take `impl CryptoRngCore` arguments,
/// this structure wraps the dynamic object and exposes a `CryptoRngCore` interface,
/// to be passed to statically typed round methods.
// TODO: remove when the RustCrypto crates are updated (#123456).
```

fjarri: Filed #216

PR #218, opened 2025-04-28
Test that `Secret` is zeroized

dvdplm:
I was looking through the source code of the `secret` crate today and saw that they have a test for zeroizing secrets. Their impl relies on an FFI call to ensure `rustc` keeps the code.

My impl is unsafe, all kinds of UB and reliant on the inner working of `SecretBox`. 

The question is: is this better than nothing?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ private-benches = ["k256", "dep:criterion"]
 k256 = ["dep:k256", "bip32?/secp256k1", "sha3"]
 bip32 = ["dep:bip32", "tiny-curve?/bip32"]
 dev = ["tiny-curve", "sha3"]
+test-allocator = []
 
 [[bench]]
 bench = true
--- a/src/tools/secret.rs
+++ b/src/tools/secret.rs
@@ -494,3 +494,56 @@ where
         self * scalar.expose_secret()
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use core::ptr;
+    use crypto_bigint::U128;
+
+    #[allow(unsafe_code, trivial_casts, unused_assignments)]
+    #[test]
+    fn clears_memory_when_scope_ends() {
+        let mut ptr: *const U128 = ptr::null();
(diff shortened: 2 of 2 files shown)


dvdplm:
> Shouldn't this rather be in the `secrecy` crate (or even `zeroize`)?

Probably, yes. I wanted to experiment with the approach first though and see just how ugly it becomes. If we like it (or doesn't dislike it tooooo much I guess), we can see if there's an interest in the `secrecy` or `zeroize` crates.

With https://github.com/entropyxyz/synedrion/pull/218/commits/05ed9b662c216b39612807b725227e63c936a3a7 I added a custom allocator that doesn't free any memory, allowing the test to inspect the "freed" after zeroization. This 'works" but has drawbacks. The stable custom allocator API only allows for a global allocator with `#[global_allocator]`. That commit "hides" the non-freeing allocator inside a module behind a feature, but that only mitigates the issue: if you run with `--all-features`, then it will activate and all tests will use the leaking allocator. This is probably not a big problem, but certainly not very elegant.

My other options are:
- Use an integration test. This requires making `Secret` pub, which we don't really want.
- Use the nightly-only `#[allocator_api]` and add a copy of the `SecretBox` code where the `Box` uses the un-freeing allocator. Eeew.
- Do what you suggest, move the test to `secrecy` or `zeroize` and see what the maintainers say about this hack.

dvdplm:
As a side-note I did some investigation on the state of the current crop of custom allocators and there are a few that are specifically geared towards handling secrets, i.e. they provide a way to zeroize all heap allocations (and some go further and use `mlock(2)`/`mprotect(2)` to stop the allocation from ever ending up in swapfiles).

When/if the `allocator_api` RFC lands, we'll have better options than the `zeroize` crate.

- [`MiMalloc`](https://lib.rs/crates/mimalloc) in secure mode (a Microsoft project)
- [`zeroizing-alloc`](https://lib.rs/crates/zeroizing-alloc) (by 1password)
- …and probably more.

PR #220, opened 2025-05-08
Statistical CT tests

dvdplm:
- **Embryo of a dudect suite of tests**
- **cleanup**
- **Add a few more tests and reorganize files a bit**
- **Renames and reorg**
- **Add more tests. Set input size to where CT breaks down for each.**
- **Tweak doc strings**


src/private_benches/zk_proofs_ct.rs
```diff
@@ -0,0 +1,113 @@
...
+    uint::{HasWide, PublicSigned, SecretSigned},
+    SchemeParams,
+};
+
+type Pai = PaillierProduction112;
+type Prm = ProductionParams112;
+
+/// Is [`RPParams::commit`] constant time?
+pub fn rp_commit_both(runner: &mut CtRunner, rng: &mut BenchRng) {
+    let (rp_params, sk, inputs) = rp_inputs(rng, /* estimate, I ran out of patience here */ 200_000);
```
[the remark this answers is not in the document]
fjarri: "I ran out of patience" -> "This value was chosen based on practical considerations". Thanks, copilot :)

examples/dudect.rs
```diff
@@ -0,0 +1,18 @@
+//! This file collects Constant Time "benchmarks" using the `dudect-bencher` crate.
```
fjarri, line 1: So what's the reason this cannot live in `benches`?

src/private_benches/CT_README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Methodology
+
+1. **Input Generation:** Two input classes are created: "normal" (random) and "special" (potentially triggering timing variations).
+2. **Timing Measurement:** Execution time of the target code is measured for each input.
+3. **Statistical Analysis:** Welch's t-test is applied to compare the execution time distributions of the two classes. The "max t" statistic is computed.
+
+### Interpretation
+
+A small absolute value of "max t" suggests that execution times for the two input classes are similar, indicating potential constant-time behavior.  A threshold of `|max t| >= 5` is often used to flag non-constant-time behavior. However, constant-time analysis can be complex, and a high "max t" value might not always indicate a practically exploitable timing leak.
```
fjarri, line 13: Is there any source for the threshold choice?

fjarri:
Looks good to me except for a couple of minor questions.

src/private_benches/secret_signed_ct.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Is `SecretSigned::new_positive` constant time when the bound is kept constant and the value varies?
+pub fn new_positive_with_constant_bound(runner: &mut CtRunner, rng: &mut BenchRng) {
+    let bound = 127;
+    let modulus = NonZero::new(U4096::ONE << bound).unwrap();
+    let input_len = 20_000_000;
+    let mut inputs = {
+        (0..input_len)
+            .map(|_| {
+                // "Normal" class: a random, typically big, value.
+                if rng.r#gen::<bool>() {
```
fjarri, line 15: Do the classes have to be interleaved randomly? Or can we generate a vector of one class and a vector of another and concatenate them?

src/private_benches/CT_README.md
dvdplm: The paper itself uses `4.5` and the `dudect-bencher` crate picks `5` without explicit justification afaict.

I think the choice of `5` as a threshold comes from several directions:

- They do multiple statistical tests (100 I think), discarding different percentiles for each test; when you do this you need stricter threshold than normal.
- It works well empirically. What actually matters here is that non-CT code will make `t` grow with the number of samples, while CT code stays constant. 5 seems to work well in practice. 

I was a bit unsure about just how much commentary to put here. Maybe this stuff is obvious to people that know some statistics? Or maybe not?

examples/dudect.rs
dvdplm: Short answer is "because it doesn't work!".

It can be an example or a binary, but when set up as a benchmark and ran with `cargo bench --bench dudectbench --features … …` it fails with an `error: Found argument '--bench' which wasn't expected, or isn't valid in this context`

I think the dudect-bencher crate is a bit old and un-loved and that if we take this on, we might have to do some work to fix a few quirks here and there.

I don't have a strong opinion between example or binary, either kind of suck.

PR #223, opened 2025-05-22
Add release-plz workflow

cooldracula:
This adds the release-plz CI using their [basic, recommended workflow yaml](https://release-plz.dev/docs/github/quickstart#3-setup-the-workflow) as a solution to https://github.com/entropyxyz/devops-infrastructure/issues/133.

I would like to iterate on this by creating a release-plz bot app and use that for our github token in the CI (see the [documented recommendation](https://release-plz.dev/docs/github/token#use-a-github-app)), but I don't think this is necessary for this first version and can follow up with a future PR.

### DIFF
--- a/.github/workflows/release-plz.yaml
+++ b/.github/workflows/release-plz.yaml
@@ -1,0 +1,59 @@
+---
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+
+  # Release unpublished packages.
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: release-plz/action@dde7b63054529c440305a924e5849c68318bcc9a # v0.5.107
+        with:
+          command: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  # Create a PR with the new versions and changelog, preparing the next release.
+  release-plz-pr:
+    name: Release-plz PR
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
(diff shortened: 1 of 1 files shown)


fjarri:
So, stupid question, what's the procedure for invoking it, and what does it actually do (that is, what needs to be done manually in addition to that)?

HCastano:
I don't think we need to do it for this repo given the development velocity, but I do want to bring up that it is possible to run `release-plz` on, for example, [a weekly schedule`](https://release-plz.dev/docs/github/advanced#run-on-schedule), instead of every commit.

Could come in handy in the future or in other repos.

HCastano:
> So, stupid question, what's the procedure for invoking it, and what does it actually do (that is, what needs to be done manually in addition to that)?

@fjarri As it is configured right now, it'll open a PR by itself every time a new change is merged into the `master` branch. When that PR is approved and merged it'll go ahead and publish all unreleased crates.

dvdplm:
> Wait, it'll open a PR after _every_ other PR, just on the off chance we want to make a release? Isn't that a little too many?

If one is open it'll just keep adding commits to it until it's merged, at least that's how I understood it. In essence there'll always be an open release PR, but nothing is released until that PR is merged.

cooldracula:
> > @cooldracula Is there an easy way to setup a crates registry that we can use to toy with this stuff? If yes, then I think we should make a (private) duplicate of this repo and point it to our own crates registry so we can test this out properly without publishing a bunch of crate versions we don't actually want to publish.

It would be relatively simple to set up a local private registry using something like [cratery](https://github.com/cenotelie/cratery).  You'd then [add this registry to your crates config](https://doc.rust-lang.org/cargo/reference/registries.html#using-an-alternate-registry) and update the Cargo.toml to use it.
However, it seems simpler to me to have a test repo with a basic rust package that we publish to crates.io and do the experiments in that repo's code?

dvdplm:
> However, it seems simpler to me to have a test repo with a basic rust package that we publish to crates.io and do the experiments in that repo's code?

I would probably just duplicate `synedrion`, `manul` and `crypto-primes` into private repos, and then do our experiments there. I tried the basic mechanics of `release-plz` on my own toy crate but the real test is on code with actual CI pipelines, and cross-repository dependencies etc.

If that's too much overhead we can also just rip the band-aid and just do it for all three repositories and just iterate on it "live". Downside is, as noted, that we'll only discover issues slowly over time as we make releases.

@fjarri What are your thoughts on this?

dvdplm:
> by simply creating a cross-repository dependency in some example code?

I'd probably just duplicate the repos and do it that way. I think we're suggesting the same thing, but to me, setting up a set of toy repos and example code + configuring CI to be a reasonable replica of our production setup seems like more work.

fjarri (approved):

PR #224, opened 2025-06-17
Use `derive-where` instead of explicit bounds to derive `serde` traits

### LINKED ISSUE #10: Can we get rid of `serde(bound)` annotations?

fjarri:
Fixes #10

Also pin the trunk version of `manul` since it requires some fixes to work around a bug in `derive-where` (see https://github.com/ModProg/derive-where/pull/122)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,8 @@ readme = "README.md"
 categories = ["cryptography", "no-std"]
 
 [dependencies]
-manul = "0.2"
+#manul = "0.2"
+manul = { git = "https://github.com/entropyxyz/manul.git", rev = "d23fcb8cc9e88729a55006874c464d918197a965" }
 signature = { version = "2", default-features = false }
 rand_core = { version = "0.6.4", default-features = false }
 digest = { version = "0.10", default-features = false, features = ["alloc"] }
@@ -23,6 +24,8 @@ rand_chacha = { version = "0.3", default-features = false }
 rand = { version = "0.8", default-features = false }
 elliptic-curve = { version = "0.13", default-features = false }
 ecdsa = { version = "0.16", default-features = false, features = ["signing", "verifying"] }
+derive-where = { version = "1.5", features = ["serde"] }
+
 criterion = { version = "0.5", optional = true }
 
 # Note: `alloc` is needed for `crytpto-bigint`'s dependency `serdect` to be able
@@ -42,7 +45,8 @@ sha3 = { version = "0.10", optional = true, default-features = false }
 dudect-bencher = { version = "0.6", optional = true }
 
 [dev-dependencies]
-manul = { version = "0.2", features = ["dev"] }
+#manul = { version = "0.2", features = ["dev"] }
+manul = { git = "https://github.com/entropyxyz/manul.git", rev = "d23fcb8cc9e88729a55006874c464d918197a965", features = ["dev"] }
 serde_assert = "0.8"
 tokio = { version = "1", features = ["rt", "sync", "time", "macros"] }
 rand = { version = "0.8", features = ["getrandom"] }
--- a/src/entities/full.rs
+++ b/src/entities/full.rs
@@ -8,7 +8,6 @@ use core::fmt::Debug;
 use ecdsa::VerifyingKey;
 use manul::{protocol::PartyId, session::LocalError, utils::SerializableMap};
 use rand_core::CryptoRngCore;
-use serde::{Deserialize, Serialize};
 
 use crate::{
     curve::{secret_split, Point, Scalar},
@@ -20,8 +19,8 @@ use crate::{
 };
 
 /// The result of the KeyInit protocol.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(bound(deserialize = "I: for<'x> Deserialize<'x>"))]
+#[derive(Debug, Clone)]
+#[derive_where::derive_where(Serialize, Deserialize)]
 pub struct KeyShare<P, I>
 where
     P: SchemeParams,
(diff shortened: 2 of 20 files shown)


Cargo.toml
```diff
@@ -11,7 +11,8 @@ readme = "README.md"
 categories = ["cryptography", "no-std"]
 
 [dependencies]
-manul = "0.2"
+#manul = "0.2"
+manul = { git = "https://github.com/entropyxyz/manul.git", rev = "d23fcb8cc9e88729a55006874c464d918197a965" }
```
dvdplm, line 15: Did you leave the commented line here on purpose? A comment would be good to give us a reminder on what issue needs to be solved for us to move back to the released version.

Cargo.toml
```diff
@@ -42,7 +45,8 @@ sha3 = { version = "0.10", optional = true, default-features = false }
 dudect-bencher = { version = "0.6", optional = true }
 
 [dev-dependencies]
-manul = { version = "0.2", features = ["dev"] }
+#manul = { version = "0.2", features = ["dev"] }
+manul = { git = "https://github.com/entropyxyz/manul.git", rev = "d23fcb8cc9e88729a55006874c464d918197a965", features = ["dev"] }
```
dvdplm, line 49: Same comment as above.

src/entities/full.rs
```diff
@@ -20,8 +19,8 @@ use crate::{
 };
 
 /// The result of the KeyInit protocol.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(bound(deserialize = "I: for<'x> Deserialize<'x>"))]
+#[derive(Debug, Clone)]
+#[derive_where::derive_where(Serialize, Deserialize)]
```
dvdplm, line 23: What does this actually expand to? Like, does it expand to exactly the same thing as what `#[serde(bound(deserialize = "I: for<'x> Deserialize<'x>"))]` expands too? Or does it add all items in the `where` clause?

Cargo.toml
fjarri: No, just an omission.

fjarri:
The answer to https://github.com/entropyxyz/synedrion/pull/224#discussion_r2176778800 is pretty lengthy, so I'd rather put it here at the top level.

As far as I understand how `derive-where` works, it just doesn't add any bounds to the `impl`s, relying on whatever the user explicitly specified in the `where` clauses for the struct or in the type bounds of the associated types of the traits used. 

From the `serde` side, there are actually several issues that cause problems. One is related specifically to deserialization. Consider this struct
```rust
#[derive(Deserialize)]
pub struct Foo<I>(I);
```

`serde` generates:
```rust
impl<'de, I> _serde::Deserialize<'de> for Foo<I> where I: _serde::Deserialize<'de>,
```

But if we already have a `Deserialize` bound on `Foo` (coming from `PartyId` in our case):
```rust
trait PartyId: for<'de> Deserialize<'de> {}
#[derive(Deserialize)]
pub struct Foo<I: PartyId>(I);
```
`serde` still adds one, and it causes a clash:
```rust
impl<'de, I: PartyId> _serde::Deserialize<'de> for Foo<I> where I: _serde::Deserialize<'de>,
```
```
error[E0283]: type annotations needed: cannot satisfy `I: Deserialize<'_>`
  --> src/main.rs:11:12
   |
11 | pub struct Foo<I: PartyId>(I);
   |            ^^^^^^^^^^^^^^^
   |
note: multiple `impl`s or `where` clauses satisfying `I: Deserialize<'_>` found
  --> src/main.rs:9:10
   |
9  | #[derive(Deserialize)]
   |          ^^^^^^^^^^^
10 | //#[derive_where::derive_where(Deserialize)]
11 | pub struct Foo<I: PartyId>(I);
   |                   ^^^^^^^
note: required for `Foo<I>` to implement `Deserialize<'de>`
  --> src/main.rs:9:10
   |
9  | #[derive(Deserialize)]
   |          ^^^^^^^^^^^ unsatisfied trait bound introduced in this `derive` macro
10 | //#[derive_where::derive_where(Deserialize)]
11 | pub struct Foo<I: PartyId>(I);
```
Using `derive_where(Deserialize)` instead gets rid of the additional `where`, leaving only the `PartyId` bound, which works fine.

Come to think of it, it may be a compiler bug: if `I: for<'de> Deserialize<'de>`, shouldn't it follow that `I: Deserialize<'de>`? Filed https://github.com/rust-lang/rust/issues/143304, but it may be a duplicate. I guess the answer will be that it will be resolved by the new type solver, at some very distant point in time. 

There is another issue with `serde` producing an incorrect bound for the type that's not used in the structure (only its associated type is), which may be https://github.com/serde-rs/serde/issues/2463, I am trying to make an MRE now.

fjarri:
Yes, the second issue is exactly https://github.com/serde-rs/serde/issues/2463, and again it's resolved by simply not adding any additional constraints. MRE:
```rust
trait MyTrait {
    type MyType: Serialize;
}

#[derive(Serialize)]
struct Container<T: MyTrait>(Inner<T>);

#[derive(Serialize)]
struct Inner<T: MyTrait>(T::MyType);

fn bar<T: Serialize>(_x: &T) {
    unimplemented!()
}

fn test<T: MyTrait>(x: Container<T>) {
    bar(&x);
}
```

`serde` produces
```rust
impl<T: MyTrait> _serde::Serialize for Container<T> where T: _serde::Serialize,
```
while `derive_where` just produces 
```rust
impl<T: MyTrait> _serde::Serialize for Container<T>
```

PR #226, opened 2025-07-23
Update to static rounds

### LINKED ISSUE #216: Remove `BoxedRng` when all the dependencies take `impl CryptoRngCore + ?Sized` arfuments
This is a temporary adapter that we need to accommodate some dependencies (specifically, `elliptic-curve` and `ecdsa`). `manul`'s trait methods receive a `dyn CryptoRngCore`, so we need to wrap it in a static type before passing it to the lower levels.

fjarri:
Use static rounds from https://github.com/entropyxyz/manul/pull/117 and helpers from https://github.com/entropyxyz/manul/pull/113

Also fixes #216 (static round methods use `impl CryptoRngCore`)


src/protocols/aux_gen.rs
```diff
@@ -222,63 +131,138 @@ impl<P: SchemeParams, Id: PartyId> ProtocolError<Id> for AuxGenError<P, Id> {
...
+                let r3_dm = messages.direct_message()?;
+                let paillier_pk = r2_nb.paillier_pk.into_precomputed();
                 let rp_params = r2_eb.rp_params.to_precomputed();
                 verify_that(!r3_dm.psi.verify(&paillier_pk, &rp_params, &aux))
             }
         }
     }
 }
 
+/// Reconstruct `rid` from echoed messages
```
dvdplm, line 246:
```suggestion
// Reconstruct `rid` from echoed messages
```

src/protocols/aux_gen.rs
```diff
@@ -222,63 +131,138 @@ impl<P: SchemeParams, Id: PartyId> ProtocolError<Id> for AuxGenError<P, Id> {
...
+fn reconstruct_rid<P: SchemeParams, Id: PartyId>(
+    messages: &EvidenceMessages<'_, Id, Round3<P, Id>>,
+) -> Result<BitVec, EvidenceError> {
+    let r2_ebs = messages.combined_echos::<Round2<P, Id>>(2)?;
+    let r2_eb = messages.previous_echo_broadcast::<Round2<P, Id>>(2)?;
```
dvdplm, line 251: We discussed this a fair bit during the `manul` review, but I nonetheless was surprised to see that we have to pass in the `2` here. It's a wart. :/

src/protocols/interactive_signing.rs
```diff
@@ -47,6 +39,35 @@ use crate::{
     },
 };
 
+/// Analogous to `Iterator::sum()`, but requires a non-empty iterator
+/// (so that it can be used for types with no `default()`, like `Ciphertext`)
```
dvdplm, line 43: Not sure I understand this comment about `Default`. Wouldn't an impl Sum for CipherText work?

```
impl<P: PaillierParams> Sum for Ciphertext<P> {
    fn sum<I: Iterator<Item = Self>>(mut iter: I) -> Self {
        let mut result = iter.next().unwrap();
        for item in iter {
            result = result + item;
        }
        result
    }
}
```

dvdplm (approved):
LGTM

src/protocols/aux_gen.rs
fjarri: The consequence of one round type being able to be used for several actual rounds.

dvdplm: Yeah I understand it, but still had to pause. It's just a bit of an oddity. Maybe if rounds had "functional" names it'd be less jarring? Like, dunno, "KeyInitRound", "ReshareRound", etc

fjarri:

src/protocols/aux_gen.rs
fjarri: I intended it to be a doc comment, I think that's fine?

src/protocols/interactive_signing.rs
fjarri: It'll panic for an empty iterator, and the purpose of these functions is to have a `Result` instead. We can't return `Result` from `sum()`.

